PR #4, opened 2015-11-17
Automatically set srcSet attributes

ekosz:
This adds a new option called `srcSet` which defaults to true. When enabled, the 2x and 3x srcSet attributes will be added to rendered component. Currently, the attribute will only be added if the rendered component is an `img`.

What do you all think? I don't know if the option name is the best at the moment.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,7 +35,8 @@ export default class ReactImgix extends Component {
     fluid: PropTypes.bool,
     children: PropTypes.any,
     customParams: PropTypes.object,
-    entropy: PropTypes.bool
+    entropy: PropTypes.bool,
+    generateSrcSet: PropTypes.bool
   }
   static defaultProps = {
     precision: 100,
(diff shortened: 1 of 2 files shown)


kellysutton:
Two cents on the naming: the flag should likely be named `includeSrcset` or `generateSrcset` as to not be confused with an actual `srcset` value. What do you think?

ekosz:
Yep! I was originally going to do that, but it didn't seem to matchup with the other options naming-wise. I can make the change now. Also do you think this option should be enabled by default?

kellysutton:
Because `srcset` is a valid `` attribute, I think it will need a prefix like `include` or `generate`.

Because `srcset` merely upgrades the functionality of an image element, there's no harm in including it even when it is not supported. Thus, I think it should be enabled by default.

ekosz:
@kellysutton Alright, I've updated the PR with the changes.

PR #6, opened 2015-12-01
make style overridable for bg={true}

athibaud:
for cases where you need to override `backgroundSize`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -103,9 +103,9 @@ export default class ReactImgix extends Component {
         component = 'div'
       }
       childProps.style = {
+        backgroundSize: 'cover',
         ...childProps.style,
-        backgroundImage: `url(${src})`,
-        backgroundSize: 'cover'
+        backgroundImage: `url(${src})`
       }
       delete childProps.src
       delete childProps.srcSet
--- a/src/index.test.js
+++ b/src/index.test.js
@@ -51,6 +51,33 @@ describe('background mode', () => {
     expect(vdom.props.style.backgroundSize).toBe('cover')
   })
 })
+describe('background mode + style prop', () => {
+  const backgroundImage = 'url("https://assets.imgix.net/unsplash/bridge.jpg")'
+  beforeEach(() => {
+    tree = sd.shallowRender(
+      <Imgix
+        src={src}
+        bg
+        aggresiveLoad
+        style={{
+          backgroundImage,
+          backgroundSize: 'contain',
+          top: 100
+        }}
+      />
+    )
+    vdom = tree.getRenderOutput()
+    instance = tree.getMountedInstance()
+  })
+  it('should render a div', () => {
+    expect(vdom.type).toBe('div')
+  })
+  it('should have the appropriate styles', () => {
+    expect(vdom.props.style.backgroundImage).toInclude(src)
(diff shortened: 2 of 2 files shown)


kellysutton:
Hi @athibaud,

Thanks for opening a PR. At first glance though, it doesn't look like this code accomplishes what you are going after. Can you please write a few new tests to encapsulate the new logic you were hoping to introduce, and add them to the PR?

athibaud:
@kellysutton 
> Thanks for opening a PR. At first glance though, it doesn't look like this code accomplishes what you are going after.

i want to make all style props overridable except `backgroundUrl` as well as maintain `backgroundSize: 'cover'` as the default if not overridden.

> Can you please write a few new tests to encapsulate the new logic you were hoping to introduce, and add them to the PR?

sure, will do

athibaud:
@kellysutton i added some tests :-)

frederickfogerty:
@athibaud I think `background-image` should be overwriteable as well, if the developer sees fit. Otherwise I think this is a good catch :+1:

athibaud:
@frederickfogerty seeing this is an image component and the `src` prop is required i didn't think it was wise to let one overwrite `background-image` but that's your call :-).

nevertheless, i think a common use case would be providing a default `background-image` via the style prop or via external css and if `src` is not set or is null.. not set the `style.backgroundImage` prop (which  would currently be set to `url()`) so that it doesn't override the *default* `background-image`.

i'll can fix code and tests + rebase

frederickfogerty:
If that is a common use case I would suggest it should be a `defaultBackgroundImage` prop on the element. Imo, if I declare something like `backgroundImage` as a prop, I want it to override anything.

It's kind of like the (undocumented) `defaultWidth` and `width` props. If `defaultWidth` is set, it uses that if it can't calculate the width or fluid layout is turned off. But, if width is set, it overrides anything that the component calculates. https://github.com/jnuine/react-imgix/blob/master/src/index.js#L19

athibaud:
@frederickfogerty i see where you're coming from but i'm not convinced adding more props is a good idea. having so many props options seems like bad design and obviously complicates the components' logic.

my intuition is it would be better to have simpler/stricter ~userspace/sugared components like `Imgix.Img` and `Imgix.BackgroundImg` components which would both have `defaultSrc` and `src` props.
similarly, (to bring back up the coversation in [here](https://github.com/imgix/react-imgix/pull/5#issuecomment-160558582)) having a `Imgix.Fluid` component could make sense.

you could still expose a default *big* `Imgix`which has all possible options for more specific needs.
and even implement the ~userspace components on top of that big `Imgix`component with something like [recompose](https://github.com/acdlite/recompose).

but thats drifting away from the issue! :-)
i'll open a conversation issue for this when i have time and clearer ideas :-)

in the meantime i think what you propose is the easier for now. ie, let the user overwrite all style props, including `style.backgroundImage`

```diff
       childProps.style = {
-        ...childProps.style,
         backgroundImage: `url(${src})`,
-        backgroundSize: 'cover'
+        backgroundSize: 'cover',
+        ...childProps.style
       }
```

i can implement that, fix tests, rebase

frederickfogerty:
Closing due to inactivity

PR #9, opened 2016-01-29
Better child props

frederickfogerty:
Change child props behaviour to only pass down props not used, not every prop

- as a bonus, we no longer have to delete child props!


frederickfogerty:
When #10 is merged, this can be rebased and the tests will pass

paulstraw:
Happy to handle this one now that #10 is merged, if you'd like.

frederickfogerty:
@paulstraw just rebased, tests should pass

paulstraw:
Nifty, thanks @frederickfogerty. :grin:

PR #10, opened 2016-01-29
Babel 6

frederickfogerty:


paulstraw:
Nice! I'll take a look at this in a bit.

paulstraw:
Oh, never mind. Super straightforward. LGTM as soon as the tests pass!

frederickfogerty:
So looks like node versions <4 are failing.. maybe I need to add a polyfill? @paulstraw

paulstraw:
Taking a peek at this now.

paulstraw:
Using `transform-object-assign` makes everything happy. I'm good with it if you are, @frederickfogerty.

frederickfogerty:
I’m guessing that’s included in the es2015 preset right? Then why are the tests failing..

paulstraw:
It's not included, no. It's a separate plugin. I added it in 4034663, which makes old runtimes work as expected.

frederickfogerty:
Ah - I was on mobile and didn't see you had committed :smile: looks good to me!

PR #15, opened 2016-02-22
expose updateImageState to parent components

theolampert:
adds a separate updateImageState to allow a parent component to trigger an image reload

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -54,14 +54,20 @@ export default class ReactImgix extends Component {
     height: null,
     mounted: false
   };
-  componentDidMount = () => {
+
+  forceLayout = () => {
     const node = ReactDOM.findDOMNode(this)
     this.setState({
       width: node.scrollWidth,
       height: node.scrollHeight,
       mounted: true
     })
-  };
+  }
+
+  componentDidMount = () => {
+    this.forceLayout()
+  }
+
   _findSizeForDimension = dim => findSizeForDimension(dim, this.props, this.state);
 
   render () {


paulstraw:
This seems fine to me, but I'm not an expert in React idioms. @frederickfogerty?

frederickfogerty:
@theolampert 

What are your goals with this change? I'm assuming you want it to recalculate the styles based on a change in width/height? To cover that use case, I'm looking at (and would prefer) the use of [DOM Mutation Observers](https://developer.mozilla.org/en/docs/Web/API/MutationObserver), which pick up these changes without having to call `updateImgState` directly.

Am I right in thinking the intended api for this is `this.refs.Imgix.updateImgState()`? If it's for the use case described above, I think a name like `forceLayout()` or `recalculateStyles()` would be much more fitting.

In terms of whether I think this is an acceptable PR, I think yes. Even though we will implement Observers (as above), I think this is still necessary, perhaps for older browsers or other scenarios.

Please let me know if I'm completely on the wrong course.

theolampert:
@frederickfogerty  yes you're correct - I ran into an issue at work where we needed this, sorry for not being clearer above. I think `forceLayout()` works well, I'll update the pr if you're happy with that.

frederickfogerty:
Okay cool. If you can squash your commits with a useful commit message I'm happy to merge. There's going be some issues here with unmounted components but I'm going to open a seperate issue for that, as it's not directly related to this PR.

theolampert:
@frederickfogerty commits are squashed and renamed now.

frederickfogerty:
Merged and released in `v2.2.0`

PR #20, opened 2016-03-05
Ensure all query keys + B64 variants are encoded.

paulstraw:
Hey @frederickfogerty, it'd be great if you could get take a quick peek at this. :smile:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -62,11 +62,11 @@ export default class ReactImgix extends Component {
       height: node.scrollHeight,
       mounted: true
     })
-  }
+  };
 
   componentDidMount = () => {
     this.forceLayout()
-  }
+  };
 
   _findSizeForDimension = dim => findSizeForDimension(dim, this.props, this.state);
 
--- a/src/support.js
+++ b/src/support.js
@@ -6,6 +6,7 @@ Minor syntax modifications have been made
 */
 
 var Uri = require('jsuri')
+var Base64 = require('js-base64').Base64
 
 // @see https://www.imgix.com/docs/reference
 var PARAM_EXPANSION = Object.freeze({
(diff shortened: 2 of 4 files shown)


paulstraw:
Hah, well uh… I guess that's… working. 👀

src/support.js
```diff
@@ -126,15 +127,20 @@ function processImage (src, longOptions) {
...
+  Object.keys(longOptions).forEach(function (key) {
+    var val = longOptions[key]
+
+    if (PARAM_EXPANSION[key]) {
+      key = PARAM_EXPANSION[key]
     }
+
+    key = encodeURIComponent(key)
+
+    if (key.substr(-2) === '64') {
```
ekosz: While `encodeURIComponent` is probably a good idea to add, `js-base64` (a 6.6kb dep) and these next few lines seem very user-landy. As in it doesn't seem to be the most general case that most or many users of this library will get use from.

frederickfogerty:
Yeah I'm with @ekosz on this one, I'm not sure of the motive for this.. don't browsers do this by default anyway?

paulstraw:
Unfortunately, the built-in `atob` and `btoa` browser methods do not output URL-formatted Base64 (which is easy enough to shim with a `String.prototype.replace`), and more importantly don't support any characters outside Latin-1 (which requires a full reimplementation). js-base64 is the lightest, best tool I've found to do this.

Auto Base64 encoding is something that we're in the process of adding to every imgix library, and should be getting very significant usage starting soon.

src/index.js
```diff
@@ -62,11 +62,11 @@ export default class ReactImgix extends Component {
       height: node.scrollHeight,
       mounted: true
     })
-  }
+  };
```
frederickfogerty: Not needed with latest babel :)

frederickfogerty:
You're going to put emojis in URLs? That's pretty exciting.

I guess it's not too bad. Maybe I'd like to see a build without base64 encoding though, maybe something like:

```js
import Imgix from 'react-imgix' // No base64 encoding
import Imgix64 from 'react-imgix/base-64'

import b64 from 'js-base64'
Imgix.setBase64Library(b64) // Allow base64 dependency injection
```

src/index.js
paulstraw: Hmm, that doesn't seem to be the case in 6.4.5, which is the version react-imgix is currently on. Even on a clean checkout, `npm run test` gives:

```
SyntaxError: /Users/paulstraw/dev/imgix/react-imgix/src/index.js: A semicolon is required after a class property (65:3)
  63 |       mounted: true
  64 |     })
> 65 |   }
```

paulstraw:
I'm pretty hesitant to start splitting things up like that. Since js-base64 is only ~6KB unminified/gzipped (and ~1.5KB minified+gzipped). I'm happy to pay that price in order to make sure that all consumers get a consistently good experience, especially given that we're actively promoting this feature with our user base. You cool if I merge this, @frederickfogerty? Happy to reevaluate later if it causes issues.

frederickfogerty:
Yup that sounds fine. Webpack/Rollup will dedupe if the consumer uses it anyway.

src/index.js
frederickfogerty: Might need to do the old upgrade... `^6.4.5` should match `^6.5` though, which is what I'm on personally

Issue #21, opened 2016-03-14
Thoughts on option to strip protocol from urls.


tryggvigy:
Many services return image urls with `http://` as the protocol. If requested over `https://` they will result in errors.

A boolean prop that would strip the protocol from the `src` attribute would be useful for many developers.
```jsx
<Imgix
  src={imageUrl}
  stripProtocol
/>
```

Would you be interested in a PR implementing this?

frederickfogerty:
@tryggvigy yup! I would accept a PR for that.

I've also thought about adding something like `domain` and then `route` or `image` so the user wouldn't have to worry about joining urls etc. What do you think?

tryggvigy:
@frederickfogerty Thanks for the swift reply!
I think string concatenation is less of an issue for devs especially with template strings.
On the other hand removing specific parts of a string (e.g. protocol) is more off a hassle.

At least thats my opinion! Keeping the API surface small is awesome. :smile: 

With your blessing I will start work on a PR implementing `stripProtocol` 
attribute as soon as possible.

frederickfogerty:
Closing as discussion developed in #22

PR #22, opened 2016-03-15
Strip protocol

tryggvigy:
### Purpose

This PR implements the `stripProtocol` property that was discussed in https://github.com/imgix/react-imgix/issues/21. If it's truthy, the component will strip the protocol from the `src` attribute passed to it.

### Example
```jsx
<Imgix
  src={'https://mysource.imgix.net/demo.png'}
  stripProtocol
/>
```
would transform the url string into  `//mysource.imgix.net/demo.png`

The specific protocol is not checked so `foobar://mysource.imgix.net/demo.png`
would also be stripped to `//mysource.imgix.net/demo.png` and make a valid request.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,10 +82,10 @@ export default class ReactImgix extends Component {
       customParams,
       entropy,
       faces,
-      src,
-      stripProtocol,
       fit,
       generateSrcSet,
+      src,
+      stripProtocol,
       ...other
     } = this.props
     let _src = ''
--- a/src/index.js
+++ b/src/index.js
@@ -31,6 +31,7 @@ export default class ReactImgix extends Component {
     fit: PropTypes.string,
     auto: PropTypes.array,
     faces: PropTypes.bool,
+    stripProtocol: PropTypes.bool,
     aggresiveLoad: PropTypes.bool,
     fluid: PropTypes.bool,
     children: PropTypes.any,
@@ -44,6 +45,7 @@ export default class ReactImgix extends Component {
     fluid: true,
     aggresiveLoad: false,
     faces: true,
+    stripProtocol: false,
     fit: 'crop',
     entropy: false,
     auto: ['format'],
@@ -80,11 +82,12 @@ export default class ReactImgix extends Component {
       customParams,
       entropy,
       faces,
+      stripProtocol,
       fit,
       generateSrcSet,
-      src,
       ...other
     } = this.props
+    let src = this.props.src
     let _src = ''
     let srcSet = ''
     let _component = component
@@ -109,6 +112,10 @@ export default class ReactImgix extends Component {
         width,
         height
       }
+
+      // Match first instance of <anything>:// in the string
+      const protocolMatch = new RegExp(/^.*?:\/\//)
+      if (stripProtocol) src = src.replace(protocolMatch, '//')
 
       _src = processImage(src, srcOptions)
       const dpr2 = processImage(src, {...srcOptions, dpr: 2})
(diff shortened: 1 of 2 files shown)


ekosz:
BTW, this is now considered an anti-pattern by many (including me): http://www.paulirish.com/2010/the-protocol-relative-url/

Not that I have a stick in this game, but for some reason the bot keeps asking for my opinion.

ekosz:
Other than that the code looks good :+1:

tryggvigy:
Good point, maybe a `forceHttps` attribute would solve the same inconvenience in a better way.
Thoughts?

ekosz:
Hmm, I would think I nicer API would be to have `Imgix` take two props: `sourceName` and `path`. 

`sourceName` would be the name of the imgix subdomain you're using, and `path` would be the path of the file you want to transform.

If those two properties are given, then `src` can be dropped and `Imgix` will construct the `src` url for you (defaulting to HTTPS of course).

So the code would look like:

```javascript
let src;
if (this.props.src) {
  src = this.props.src;
} else if (this.props.sourceName && this.props.path) {
  src = `${this.props.sourceName}.imgix.net/${path}`;
} else {
  throw new Error("Imgix requires a 'src' prop, or 'sourceName' and 'path'");
}
```

paulstraw:
That's more in line with how other imgix libraries work, and seems like a nice compromise to me.

tryggvigy:
@frederickfogerty was discussing exactly this in #21.
On second thought this looks like a superior solution.

frederickfogerty:
See #21.

@ekosz that's interesting that `//` has become an anti-pattern. Of course we can't make all urls HTTPS as some servers don't support HTTPS, but for imgix this will be fine.

tryggvigy:
`forceHttps` option could be very handy in some cases.

For example we get a full `http://` imgix url from a 3-rd party server which we have no control over and the `sourceName` and `path` would have to be fished out of the string.
It would be very convenient if the component could force `https://` for us.

frederickfogerty:
Yes... I'm still unsure about all of this.

I reckon all of these props could be implemented in user-land quite easily:

```js
import Imgix, { forceHTTPS, stripProtocol, createUrlWithDomain } from 'react-imgix'

const coverImage = ({ url }) => {
  const HTTPSurl = forceHTTPS(url)
  const urlNoProtocol = stripProtocol(url)
  const fullURL = createUrlWithDomain({ domain: 'facebook', route: '/cover.png' })
  return <Imgix src={ } />
}
```

Many libraries, for example react-motion, and redux, have implemented this to much success. I believe it is better for the following reasons:

- reduces API surface area (watch this [talk for why](https://www.youtube.com/watch?v=4anAwXYqLG8))
- lets core focus on its core value add, which is the the auto-resizing of images based on element dimensions
- reduces complexity in core.

I'm trying to do this in other libraries I contribute to (see [here](https://github.com/philcockfield/ui-harness/issues/42) and [here](https://github.com/philcockfield/ui-harness/issues/31)).

Keen to hear your thoughts @ekosz, @tryggvigy and @paulstraw.

ekosz:
:+1: I'm all for small API services. You're absolutely correct that the url transforming is a user land feature.

Here's the code that can be used to do the force-https BTW: http://stackoverflow.com/a/5491311/127629

tryggvigy:
> url transforming is a user land feature
Yes. Leaning towards this.

@frederickfogerty this is exactly what we ended up doing. 
I guess I thought of it as an inconvenience the component could handle for me easily.

I see no real benefit in this PR being merged, 
in the end it really is not that much of a hassle to transform a url.
Closing PR.

PR #26, opened 2016-05-20
Fix #25

### LINKED ISSUE #25: Fluid on one dimension
I'm trying to make the image fluid on only one dimension, but this library seems to not allow me to do it.

One way to resolve it I think is to support float height/width (https://docs.imgix.com/apis/url/size/h). Right now if I pass the props of `height = 1.0` to the component, it outputs an image literally with height of 1px, which is behaving differently from the imgix documentation.

minfawang:
This fix prevents the `width` and `height` passed down as props to the rendered element. So the dimensions of the `img` tag will not be overwritten unexpectedly. See #25 for more details.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -116,7 +116,11 @@ export default class ReactImgix extends Component {
       srcSet = `${dpr2} 2x, ${dpr3} 3x`
     }
 
-    let childProps = other
+    let childProps = {
+      ...other,
+      width: null,
+      height: null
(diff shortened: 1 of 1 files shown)


paulstraw:
This looks good to me @minfawang. Would you mind possibly adding a test to verify that it works as expected?

minfawang:
Close this one in favor of #27.

PR #35, opened 2016-07-21
Add semantic release 🚀

frederickfogerty:
This adds [semantic-release](https://github.com/semantic-release/semantic-release), freeing us from having to publish ourselves. Explanation (much better than I can write) is over at their repo. Short pro/con list:

👍  Stops us having to release and publish ourselves
👍  Easy semver
👎  Forces everyone to have to write in the [angular format](https://github.com/conventional-changelog/conventional-changelog-angular/blob/master/convention.md). We can add this to the PR template so we can help them out a bit.

I'm not too fussed about this. If you like it, @paulstraw, merge it, otherwise don't.


paulstraw:
Very interesting. I'm open to this. Have you used it on other repos? Dig it?

frederickfogerty:
@paulstraw Haven't used it yet. Also, it seems like it's not being maintained any more? Sort of? Look at the commit history on the main repo 😕

paulstraw:
Thought about this a bit, and I'm down to give it a try. Worst case if there's something we don't like about it, we can revert. Sound good?

Issue #49, opened 2016-09-28
Whole HTML page can be unexpectedly requested multiple times


andykent:
There is a pretty epic and hard to spot bug in ReactImgix. It took me several hours to track this down as it's very subtle.

We were observing multiple requests to our backend servers for each pageview that included ReactImgix on it. It turns out that there is a subtle browser behaviour where a blank URL is interpreted as 'the current page' this means setting `backgroundImage = url()` will actually attempt to request the entire html page as an image. I believe this is also true of `<img src="" />` nodes too.

The two specific bugs in ReactImgix are...
- setting `bg=true` causes the initial render pass to generate styles such as `backgroundImage = url()`
- without `aggressiveLoad` set the initial render pass generates an img tag like `<img src="" srcset="" />`

It's particularly noticeable when using server side rendering as this first render pass is what gets returned to the client because there is no DOM to mount. 

For the `bg=true` case I think this can be easily fixed by simply not setting the `backgroundImage` style attr until the src attr is computed and present. E.g. we have patched this locally using...

```
if (_src && _src.length > 0) childProps.style.backgroundImage = `url(${_src})`
```

Some possible, but as yet totally untested solutions for the `src` attr might be to...
- leave the src attribute off entirely (this would be best if it works but I've not tested it)
- set `src="about:blank"` (smallest change, not sure if it would work)
- render a `<span>` tag instead of the actual image. (probably a breaking change and not really desirable but would work)

paulstraw:
Hey @andykent, thanks so much for this thorough report. It sounds like you've already got one fix in place (which looks sensible to me), and have some good theories on the second fix. Would you like to submit a PR for this? If not, I'm totally happy to work on it, but I just wanted to check since it seems like you've got a really good handle on the problem.

frederickfogerty:
😱  The new things you learn every day!

Thanks for the bug report @andykent, it seems like you have a pretty firm grasp on the problem. I think the best approach would be to work through the possible solutions you have outlined in order. Is the patch you outlined in user-land, or did you modify the library source? If you have any code that would fix this issue in the source, please feel free to submit it as a PR and we can pick it up. If not, no worries, one of us will get onto fixing it.

andykent:
We actually fix it in userland by sub-classing the component right now. I can try and have a go at a PR tomorrow though if I get a few mins to spare.

andykent:
Fixed in #51

frederickfogerty:
[...]

PR #51, opened 2016-09-29
Avoid blank background urls and src attributes

andykent:
See https://github.com/imgix/react-imgix/issues/49 for the full context.

I've added some additional tests and they are all passing but I've not actually tried running the built code in a real project yet so I can't be sure there aren't any regressions.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -131,9 +131,9 @@ export default class ReactImgix extends Component {
       }
       childProps.style = {
         ...childProps.style,
-        backgroundSize: 'cover'
+        backgroundSize: 'cover',
+        backgroundImage: isStringNotEmpty(_src) ? `url(${_src})` : null
       }
-      if (isStringNotEmpty(_src)) childProps.style.backgroundImage = `url(${_src})`
     } else {
       if (!component) {
         _component = 'img'
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,8 @@ import processImage from './support.js'
 
 const roundToNearest = (size, precision) => precision * Math.ceil(size / precision)
 
+const isPresent = (str) => str && typeof str === 'string' && str.length > 0
+
 const defaultMap = {
   width: 'defaultWidth',
   height: 'defaultHeight'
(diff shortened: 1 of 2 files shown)


paulstraw:
Looked through this, and it makes sense to me. I'll pull down shortly to verify in a test project.

paulstraw:
Finally got a chance to try this out, and it seems to be working as expected. @andykent, have you been using this branch? Run into anything? If @frederickfogerty is good with it, I'm happy to merge this.

src/index.test.js
```diff
@@ -31,6 +31,20 @@ describe('<img> mode', () => {
...
 })
+describe('<img> mode - pre-mount', () => {
+  beforeEach(() => {
+    tree = sd.shallowRender(
+      <Imgix
+        src={src}
+      />
+    )
+    vdom = tree.getRenderOutput()
+  })
```
frederickfogerty: Is this emulating the pre-mount state because `tree.getMountedInstance()` is not present?

src/index.js
```diff
@@ -5,6 +5,8 @@ import processImage from './support.js'
 
 const roundToNearest = (size, precision) => precision * Math.ceil(size / precision)
 
+const isPresent = (str) => str && typeof str === 'string' && str.length > 0
```
frederickfogerty: Could we rename to `isStringNotEmpty`?

src/index.js
```diff
@@ -85,8 +87,8 @@ export default class ReactImgix extends Component {
       src,
       ...other
     } = this.props
-    let _src = ''
-    let srcSet = ''
+    let _src = null
+    let srcSet = null
```
frederickfogerty: 👍

andykent:
Apologies for the delay. We just had a baby girl so I was off on paternity leave but back now. I've made the requested change and also inlined the conditional as I think it makes the  behaviour more obvious to parse now. 

I'll test if out on our site today and report back shortly but I think we are good.

andykent:
Yep, this is working for us in our codebase.

@frederickfogerty are you able to re-review and merge this in when you get a moment.

It would be really great for us to be able to switch back to an official release sometime soon.

frederickfogerty:
Hi @andykent I've just added a comment to the test answering my own question. I'm going to merge this now, but can you please review that comment, and if it's wrong, I can fix that up on master.

PR #53, opened 2016-10-31
Update node versions supported

frederickfogerty:
To be honest, we should be doing browser testing with Chrome if we're really serious about this


frederickfogerty:
@paulstraw feel free to merge this if this is okay with you - shouldn't be a problem but just thought I'd check with you before this is merged.

paulstraw:
Will take a look in a few, about to head into a meeting.

PR #89, opened 2017-03-21
Background Size adjustments

dennisschaaf:
We had to specify some custom background sizing on some components with CSS. This was difficult, because the background size was directly applied to the style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,6 @@ what kind of component to render, one of `img`, `bg`, `picture`, `source`. Defau
 _DEPRECATED, use `type='bg'` instead_. whether to render the image as a background of the component, defaults to `false`.  
 _To be deprecated in v6._
 
-#### backgroundSize={string}
-Allow overriding the backgroundStyle property applied to the component when rendering a `bg`, defaults to `cover`.
-
 #### component={string}
 wrapper component to use when rendering a `bg`, defaults to `div`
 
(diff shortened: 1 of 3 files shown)


frederickfogerty:
Hey there @dennisschaaf. Thanks for the PR.

I have a suggestion, please let me know what you think:

What if we change the code from 

```js
        childProps.style = {
          ...childProps.style,
          backgroundSize: backgroundSize,
          backgroundImage: isStringNotEmpty(_src) ? `url(${_src})` : null
        }
```

to 
```
        childProps.style = {
          backgroundSize: childProps.style.backgroundSize || 'cover',
          ...childProps.style,
          backgroundImage: isStringNotEmpty(_src) ? `url(${_src})` : null
        }
```

This way, we're not introducing another prop. Do you think a `backgroundSize` prop is confusing, especially since it only works when `bg=true`?

Please let me know what you think.

dennisschaaf:
Hi @frederickfogerty,

Thanks for the quick feedback!

I agree with your approach, and I think we could also make it possible to not introduce a backgroundSize property.

If we have childProps.style override the default behaviour, then the developer has complete freedom to override the styling behaviour how he needs.

Then I will change the request to be like this:

`    childProps.style = {
          backgroundSize: 'cover',
          backgroundImage: isStringNotEmpty(_src) ? `url(${_src})` : null,
          ...childProps.style,
    }`

How does that work?

frederickfogerty:
It's probably best to do it like that @dennisschaaf.

dennisschaaf:
@frederickfogerty Added my fix, as well as two tests and documentation updates.

frederickfogerty:
LGTM 👍

frederickfogerty:
Published in `5.3.0` 🎉

Issue #41, opened 2016-09-06
Adding an alt prop to imgProps loads a low res image


kamui:
This is a very strange bug, but in Firefox and Edge 14, if I add an `alt` tag to the `imgProps` prop, the image loads at 200x100 dimensions. If I remove the alt tag, it works fine. Here's a sample of my code:

``` javascript
import Imgix from "react-imgix"

<Imgix
  imgProps={{
    alt: "Testing",
  }}
  src="https://example.com/test.jpg"
/>
```

I've tried adding data attributes too, and those work fine too.

Demo: https://github.com/bonobos/react-imgix-alt-bug

frederickfogerty:
When you say Firefox and Edge, is it just those browsers, or does it happen on all browsers (such as Chrome)?

This is very weird as [this]https://github.com/imgix/react-imgix/blob/46399c50e12d5d58a46eb7cb82cbfc4299bd0689/src/index.js#L120) is literally the only place `imgProps` is used. 

Does adding an explicit `width` or `height` stop it rendering at 200x100?

kamui:
It's only Firefox and Edge 14. It works fine on Chrome and Safari.

If I pass in an explicit width and height it does stop it rendering at 200x100.

```javascript
import Imgix from "react-imgix"

<Imgix
  imgProps={{
    alt: "Testing",
  }}
  src="https://example.com/test.jpg"
  width={500}
  height={600}
/>
```

I get `?w=500&h=600`.

frederickfogerty:
That is bizarre.

Just to clear up - is the image **rendered** at 200x100, or is that the image that's **loaded** (i.e. what is requested from Imgix) is 200x100?

frederickfogerty:
Okay, so this is a little different to I had previously thought, I'm glad we got that cleared up. I'm taking a look at the source now to see if I can see anything amiss.

kamui:
Thanks. For some more context, the original image on imgix is 5500x5500.

frederickfogerty:
So.... to be honest, I have no idea what's going on.

My hunch is that is something to do with the size of the rendered block changing when the alt tag is added. It's not really a surprise that both FF and IE are different to Chrome in this regard. I'm imagining it's something to do with the CSS being applied to the image, and also something that could probably be solved with CSS. Potentially in your stylesheet, there is a style relating to images with alt tags, causing the dimensions to change, but it could be (and probably just is) the browsers being weird.

[...]

PR #94, opened 2017-04-05
add onMounted callback with access to underlying node

kochis:
This adds the ability to register an `onMounted` callback that will be called with the underlying DOM node as an argument.

This can be useful for a number of reasons, but in our case, it's a workaround for issue https://github.com/imgix/react-imgix/issues/41. This way, we can render the image without the alt tag (which won't affect the size calculation), and append the alt attribute after mounting.


frederickfogerty:
Published as `5.4.0`.

PR #102, opened 2017-04-10
pull in prop-types from a separate module, fix sinon deprecation warning

modosc:
i tested with 15.5. and 15.4 and this change worked fine - i tried to test with 0.14.x but gave up trying to chase down the required `devDependencies` changes. if this is important we might consider setting up multiple builds to verify older version compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
 	"homepage": "https://github.com/imgix/react-imgix#readme",
 	"peerDependencies": {
 		"react": ">=0.14",
-		"react-dom": ">=0.14"
+		"react-dom": ">=0.14",
+		"prop-types": ">=15.5.6"
 	},
 	"dependencies": {
 		"js-base64": "~2.1.9",
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,8 @@
 import './array-findindex';
 
 import ReactDOM from 'react-dom';
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 import { deprecate } from 'react-is-deprecated';
 
 import processImage from './support.js';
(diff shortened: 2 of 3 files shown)


frederickfogerty:
Thanks for this update @modosc. Do you know if `prop-types` is meant to work with `0.14.x`?

If it is, then I'll release this as a patch version. If not, I'll have to release this as a breaking change.

modosc:
well - it would certainly be easier to just support 15.5 going forward (compare to [this](https://github.com/glenjamin/skin-deep/pull/87) or [this](https://github.com/airbnb/enzyme/pull/876) - in that case i'd have to update the `peerDependencies` to match.

otherwise i could spend some time adding in test support for multiple react versions, etc.

frederickfogerty:
Don't you love minor version bumps with breaking changes! /s

It's important to me that we support all `15.x` versions.

Can you please make sure I'm understanding this correctly: 

`prop-types` update: This should work with all `15.x` versions, so this PR is fine for that.

skin-deep/test update: This is different between React 15.5 and 15.4, but if we just run our tests on 15.5, then our tests will cover all 15.x versions of React.

---

So, my current thinking is to release this as a breaking change, dropping support for React 0.14.

modosc:
> prop-types update: This should work with all 15.x versions, so this PR is fine for that.
> skin-deep/test update: This is different between React 15.5 and 15.4, but if we just run our tests on 15.5, then our tests will cover all 15.x versions of React.

it may not be ideal to pull in `prop-types` to 15.4 though since it's duplicated in `React.propTypes` but i tested locally with 15.4 and it did indeed work (at least, tests passed).

frederickfogerty:
@modosc thanks for clearing that up.

Agree it's not ideal to have propTypes duplicated, but at least it works. I'll add a note on the release notes so that people are aware of this duplication.

modosc:
make sure to bump up the required deps in `package.json` to 15.5.x

frederickfogerty:
@modosc I was thinking of bumping it to `15.x`.

Also, should we list `prop-types` as a dependency? 

---

There's a bit more work I need to do on this before I can release it (e.g. remove some old deprecations), I'll get around to it this arvo.

PR #109, opened 2017-04-22
Update readme to include customParams example, closes #75

### LINKED ISSUE #75: Support for Mask API
Hey @frederickfogerty, just wondering if this supports the [imgix Mask API](https://docs.imgix.com/apis/url/mask/mask)? 

Doesn't seem to be documented in the `props`, or maybe I'm missing something?

theolampert:
Closes #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,8 @@ any other Imgix params to add to the image `src`
 _For example_:
 ```js
 <Imgix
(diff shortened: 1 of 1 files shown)


frederickfogerty:
@theolampert I just updated the styles slightly to match the standard React style. (I think you copied the imgProps example, but I updated that one too). Thanks 🙌

PR #110, opened 2017-05-04
various child fixes for type=picture

modosc:
1. don't clone children unless it's actually necessary
2. don't override child's key if it's already set
3. fix child type comparison
4. update deps


modosc:
actually please hold off - some of this doesn't do what i want it do do, i'll open a new pr

PR #116, opened 2017-07-26
fix alt tag handling

modosc:
1. pass `imgProps` down to `` tags auto-generated by `type=picture` - previously this was done as a spread assignment since `className` and `styles` were the only `imgProps` we cared about (and we wanted to set those directly on the newly created `<Imgix>` component rather than its `` child). 
2. strip `alt` props from `<picture>` and `<source`> tags since they're not allowed. 
3. update deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
-    "babel-core": "^6.24.1",
+    "babel-core": "^6.25.0",
     "babel-eslint": "^7.2.3",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-preset-es2015": "^6.24.1",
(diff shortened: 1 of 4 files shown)


modosc:
@frederickfogerty you ok if i merge this?

frederickfogerty:
@modosc sorry for the delay, must've slipped through the cracks. 

This looks great, thanks! 🙌🏿

modosc:
thanks @frederickfogerty , can you roll up a release?

frederickfogerty:
Is this a breaking change @modosc?

modosc:
> Is this a breaking change @modosc?

nope, a bugfix

frederickfogerty:
Published as `6.0.1` @modosc. Thanks again 🙌🏿

PR #125, opened 2017-11-17
chore: add es build

modosc:
hah, i was really hoping to get this in before you did the 7 bump (this is why i was upgrading deps, i wanted to make that a separate change though).

the only thing that's not 100% is `build:watch` - i don't know an easy pattern to have multiple watch processes running. not sure how much that matters though, i could add a separate `commonjs` and `es` watch task


frederickfogerty:
Thanks :) Released in `7.1.1`

PR #126, opened 2017-12-12
Document default height/width props

danielfarrell:
Also add prop types for height/width/defaultHeight/defaultWidth


frederickfogerty:
Thansk @danielfarrell! Released in `7.1.0` 🎉

PR #133, opened 2018-06-10
chore: cleanup repo (prettier, docs)

frederickfogerty:
This PR updates prettier support to be more inline with the rest of imgix's JS repos. It also updates the documentation by adding a contributing file and update the readme with some installation instructions.


frederickfogerty:
Thanks @modosc 😄

PR #143, opened 2018-06-25
test: add real browser testing with Karma and BrowserStack

frederickfogerty:
## Description

This PR adds support for real browser testing with Karma and BrowserStack. Even though this project is based on React, since it uses the DOM directly, there are discrepancies between browsers. Even if this project didn't use the DOM directly, there are still discrepancies which should be tested. This shows the importance of real browser testing in addition to unit testing. Finding and controlling these discrepancies have been instrumental in solving many issues in the past, and now it can be found and controlled in a stable way. This technique has been used to solve #41.

There have been some decisions made to reduce the number of browsers tested in order to speed up the CI time. Browsers like Chrome, Edge, and Firefox have a quick release cycle. As such, it seems silly to test the last two versions of these browsers as they will little discrepancy between them. 

At the moment, browser testing on iOS is disabled as there is an issue in the configuration between Karma and BrowserStack which is preventing these from testing successfully.

### New Feature

- [x] Update the readme


frederickfogerty:
Thanks for the review @jayeb. In fact, it's already running on CI :smile: This is a screenshot from CI:

Issue #144, opened 2018-06-26
v8 Breaking Changes Discussion


frederickfogerty:
There are some things that may be good ideas to change in the next release as a breaking change. I've created this issue to track these discussions.

### Reducing the "magic"

This library does some things by default which may not make sense for the average Joe who stumbles across this component. To reduce the number of surprises for new users, these changes could be made:

- change `faces` default to false
- remove `precision` default (or make logic smarter, as below)

### Changing the `<picture>` API to remove automatic fallback creation

See more [here](https://github.com/imgix/react-imgix/issues/90#issuecomment-400166479)

Also, it seems to make sense to expose the `media` prop for `<source>` elements as a top-level prop.

### Different components for different modes (picture, bg, img)

Currently, there are some props which only make sense for a specific mode (e.g. `generateSrcSet` only makes sense for `<img>` mode, or `component` for bg mode). Having all props accessible for all modes is confusing and error-prone. An option here is to expose different components for different modes, such as Img, Picture, and Source. 

### Reduce/consolidate props

A lot of imgix "image transform" parameters, such as fit and crop, which are exposed at the top level, could be rolled into one `imgixOptions` param. This would reduce the top-level surface area, and would make the props more understandable as these options logically fit together.

jayeb:
I am extremely in favor of all four of the things you've proposed here.

modosc:
wrt the image component change - i think that’s a place where we potentially want the magic to happen (to work around the safari bug, to remove the need to provide yet another img component, etc). in both of those situations things may appear to work fine without but end up breaking in edge cases. 

i like the idea of specific components for the various modes - that would definitely simplify the props

frederickfogerty:
Good to see support for specific components :smile: 

@modosc Would you mind going into a bit more detail about the "image component change"? I'm not sure which change is being referred to here.

PR #151, opened 2018-07-01
fix: warn the user when no <img> passed as a child to <picture> fixes #90

### LINKED ISSUE #90: width and height don't do the right thing when they differ from width and height of sources
moved from [here](https://github.com/imgix/react-imgix/issues/45#issuecomment-288206488)
> Edit: issue seems to be coming from the generated fallback  element with the Picture Support. You need to define the width/height of the image on the parent <Imgix/> element otherwise it will load the 1x1 version by default on the fallback ! (crashing Safari/mobiles if you have big images by default).

something else i noticed yesterday in my work's codebase - you not only have to have a width/height on the parent element but it also needs to be the maximum width/height of any of the `<source>` elements that may be inside of the generated `<picture>`. so this:
```
<Imgix type="picture" src={src} width={100} height={100}>
  <Imgix type="source" src={src} width={400} height={100} imgProps={{ media: '(min-width: 768px)' }} />
</Imgix>
```

will load the 400x100 image on a large screen but will only display it as 100x100 (at least in chrome and safari)

i _think_ this is a bug - in my local testing setting the provided height/width on the fallback image but changing the same attributes on the parent container to the maximum values did the right thing everywhere. working around it doesn't seem hard but it's unclear if that's the expected behavior though. should i just open a new bug to discuss this?

frederickfogerty:
BREAKING CHANGE: A fallback image will no longer be created when using react-imgix in picture mode

## Description

The fallback image support when using react-imgix in `picture` mode was causing issues with sizing in #90 as the size being passed to the root `picture` component was being used for the fallback as well. This was causing problems, and the simplest solution was to remove the fallback image creation. This is inline with the HTML spec, so anyone used to using `<picture>` elements in HTML will be at home with this. There is more discussion in #90.

A problem here is that `src` is a required prop for react-imgix, which will cause an error when using `picture` mode as a `src` now isn't going to be passed to the root component. An option we have here is to update the `src` prop to not be required, and just to warn the user when a `src` is not passed when the component is being used in `img`, `bg`, or `source`. Another option is to wait to see if we implement the splitting of components as described in #144. In either way this needs to be discussed and resolved before this PR is merged.

Also, since this is a breaking change, I'm going to merge this into the `version-8` branch, and I'll look to land all the breaking changes at once (e.g. the ones being discussed in #144).

This PR fixes #90 

### Bug Fix

- [x] Update the readme
- [x] Update or add any necessary API documentation
- [x] Add some [steps](#steps-to-test) so we can test your bug fix

## Steps to Test

Observe the changes to the unit tests in `test/unit/react-imgix.test.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,24 +158,7 @@ Using the [<picture> element](https://docs.imgix.com/tutorials/using-imgix-pictu
 </Imgix>
 ```
 
-The final `type='img'` component will be created with the options passed into the parent `<picture>` container if it's not provided so the above is equivalent to:
-
-```js
-<Imgix src={src} width={100} type="picture">
-  <Imgix
-    src={src}
-    width={400}
-    type="source"
-    imgProps={{ media: "(min-width: 768px)" }}
-  />
-  <Imgix
-    src={src}
-    width={200}
-    type="source"
-    imgProps={{ media: "(min-width: 320px)" }}
-  />
-</Imgix>
-```
+It is highly recommended that a fallback `img` be passed as the last child to a picture component so that the image will render for all screen resolutions.
 
 ## Browser Support
 
--- a/src/react-imgix.js
+++ b/src/react-imgix.js
@@ -215,7 +215,7 @@ export default class ReactImgix extends Component {
         // we need to make sure an img is the last child so we look for one
         //    in children
         //    a. if we find one, move it to the last entry if it's not already there
-        //    b. if we don't find one, create one.
+        //    b. if we don't find one, warn the user as they probably want to pass one.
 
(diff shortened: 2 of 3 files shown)


frederickfogerty:
@jayeb did you have any thoughts on the problem outlined in the description?

jayeb:
The component-splitting you brought up in #144 seems like a good idea on its own, and if it also happens to solve our problem here then that's just gravy.

frederickfogerty:
Great, thanks @jayeb. I'm going to merge this into `version-8` now, and the problem will be resolved when we implement component splitting.

PR #159, opened 2018-07-26
feat: implement responsiveness with srcSet and sizes

### LINKED ISSUE #71: Include the Imgix.js generated srcset within react-imgix library
`Imgix.js` is great at what it does, unfortunately due to the nature of the html being generated after the DOM is loaded, there is always a risk of a flicker for the user.

I was looking into using this library instead to create responsive images, but noticed that the fancy srcset generation that `Imgix.js` does isn't part of this library.

It'd be great if we can have it generate that `srcset` list here as well :)

### LINKED ISSUE #115: Support for sizes, and width descriptors
Hey @frederickfogerty, thanks for putting out this library. Somewhat connected to #71, but while trying to use to do responsive images, I noticed that the library doesn't currently support `sizes` prop for ``, or width descriptors (density descriptors seem to be supported via `dpr`) within `srcSet`.

Not sure if I missed something, or if there was a reason for that? If not, would this be something on the radar to add?

Thanks!

### LINKED ISSUE #153: Can render more eagerly in certain situations than does at the moment.
**Before you submit:**

- [x] Please search through the existing issues (both open AND closed) to see if your feature has already been discussed. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please take a moment to find out whether your idea fits with the scope and aims of the project

**Is your feature request related to a problem? Please describe.**
At the moment, if the component is used in any of the following ways:

- `width` and `height` set
- `fluid={false}`

the component still takes two render passes to fully render, when it can be completed in one.

Check whether either of the following are true and eagerly render if true:

- `width != null && height != null`
- `fluid === false`

frederickfogerty:
BREAKING CHANGE: 
- srcSet behaviour has changed to use sizes + srcSets
- `type=bg` has been removed
- the following props have been removed: aggressiveLoad, component, fluid, precision, defaultHeight, defaultWidth
- generateSrcSet has been changed to disableSrcSet

## Description

This PR fundamentally changes the approach taken to responsiveness. Previously, the image (or component) was rendered with an empty image, and the size of the resulting DOM element was measured. An image was loaded with these dimensions, ensuring that the image loaded was only as large as needed. Now, this component achieves the same behaviour by leveraging new native browser features, `srcSet` and `sizes`. This allows the browser to load an appropriate size based on the expected image size, which is calculated based on `sizes`. This has an added benefit that images can load much quicker as the browser can load them as soon as page layout has completed, or sooner (e.g. if sizes is `100vw`). Overall this change has made both the usage of the library and the internals simpler and easier to understand.

The significant amount of breaking changes makes sense due to a lot of the props no longer making sense. Background mode has also been removed as it was incompatible with the new method of responsiveness, and after discussion with @jayeb it seemed that it wouldn't be missed (based on a similar change that imgix.js made).

Picture mode still doesn't work amazingly, as there is a mis-match between what a developer would expect to set as props and what they actually have to set to make picture work as they want. This is not a problem with our library per se, but with `picture` pecularities. For example, if the fallback `img` in a `<picture>` has a `width` attribute set, *all* the other sources will also be locked to that size, even if they specify their own widths. Maybe this is common knowledge but it seems peculiar to me. I expect to address this and make it clearer for developers in a future PR.

This PR closes #153. This PR closes #115. This PR closes #71

## Checklist

Please use the checklist that is most closely related to your PR _(you only need to use one checklist, and you can skip items that aren't applicable or don't make sense)_:

- [fixing typos]()
- [documentation]()
- [bug fix]()
- [new feature]()

### New Feature

- [x] Update the readme

## Steps to Test

Observe changes to unit tests. 

It may also be worth to set up a test project and use `npm link` to check default image rendering and picture rendering.


README.md
```diff
@@ -38,53 +54,69 @@ import Imgix from 'react-imgix'
...
+```html
+<img
+	src="https://assets.imgix.net/examples/pione.jpg?auto=format&amp;crop=faces&amp;fit=crop&amp;ixlib=react-7.2.0"
+	sizes="100vw"
+	srcset="https://assets.imgix.net/examples/pione.jpg?auto=format&amp;crop=faces&amp;fit=crop&amp;ixlib=react-7.2.0&amp;w=100 100w, https://assets.imgix.net/examples/pione.jpg?auto=format&amp;crop=faces&amp;fit=crop&amp;ixlib=react-7.2.0&amp;w=200 200w,..."
+>
+```
 
-For server rendering, `aggressiveLoad` should be used. This component renders nothing on the first render as it tries to work out what the size of the container it will be rendering in, and only loads an image at the resolution required. For server rendering this will mean no image will be rendered.
+Since imgix can generate as many derivative resolutions as needed, react-imgix calculates them programmatically, using the dimensions you specify (note that the w and h params scale appropriately to maintain the correct aspect ratio). All of this information has been placed into the srcset and sizes attributes.
```
jayeb: Maybe I'm missing something, but what's up with the bit about `w` and `h` scaling properly? This isn't exemplified in the above code.

src/constants.js
```diff
@@ -0,0 +1,6 @@
+const CONSTANTS = {
+  availWidth: window.screen.availWidth,
```
jayeb: I can't find good data about browser support for this. Have we verified that `window.screen` is available on all browsers we support?

src/react-imgix.js
```diff
@@ -4,44 +4,20 @@ import ReactDOM from "react-dom";
...
-const EMPTY_IMAGE_SRC =
-  "data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=";
+import targetWidths from "./targetWidths";
+import constructUrl from "./constructUrl";
 
 const PACKAGE_VERSION = require("../package.json").version;
-
-const roundToNearest = (size, precision) =>
-  precision * Math.ceil(size / precision);
+const NODE_ENV = process.env.NODE_ENV;
```
jayeb: Interesting. How does this end up manifesting when this code is run in-browser?

src/react-imgix.js
```diff
@@ -63,43 +39,89 @@ export default class ReactImgix extends Component {
...
+      type
+    } = props;
+
+    let crop = false;
+    if (faces) crop = "faces";
+    if (entropy) crop = "entropy";
+    if (props.crop) crop = props.crop;
+
+    let fit = false;
+    if (entropy) fit = "crop";
```
jayeb: I think this should be `if (entropy || faces)`?

src/react-imgix.js
```diff
@@ -63,43 +39,89 @@ export default class ReactImgix extends Component {
...
+      ...(fixedSize && height ? { height } : {}),
+      ...(fixedSize && width ? { width } : {})
+    };
+
+    const src = constructUrl(this.props.src, srcOptions);
+
+    let srcSet;
+
+    if (disableSrcSet) {
+      srcSet = src;
```
jayeb: Hmm... the `disabledSrcSet` option doesn't actually disabled `srcset`. Is that bad? Maybe there's no good reason not to just set it anyway as we're doing here, but if the option says "don't set it" the user expectation might be that it doesn't get set. I don't feel strongly either way, but it's food for thought.

src/react-imgix.js
```diff
@@ -63,43 +39,89 @@ export default class ReactImgix extends Component {
...
+          const url = constructUrl(this.props.src, {
+            ...srcOptions,
+            width: targetWidth
+          });
+          return `${url} ${targetWidth}w`;
+        };
+        const addFallbackSrc = srcSet => srcSet.concat(`, ${src}`);
+        const _srcSet = targetWidths.map(buildSrcSetPair).join(", ");
+        const _srcSetWithFallback = addFallbackSrc(_srcSet);
+        srcSet = _srcSetWithFallback;
```
jayeb: This seems like a really convoluted way of doing this. How about just doing this?

```js
srcSet = targetWidths.map(buildSrcSetPair).concat([src]).join(", ");
```

src/targetWidths.js
```diff
@@ -0,0 +1,109 @@
...
+  const maxPxWidthRequired = MAXIMUM_SCREEN_WIDTH * MAX_DPR;
+  const _screenWidths = screenWidths(maxPxWidthRequired);
+  const allWidths = deviceWidths().concat(_screenWidths);
+
+  return uniq(allWidths).sort(function(x, y) {
+    return x - y;
+  });
+}
+
+export default targetWidths();
```
jayeb: This is an excellent update to the crusty version found in imgix.js, thank you. At some point in the near future I'd love to pull this out into its own repo with a JSON distributable a la https://github.com/imgix/imgix-url-params.

src/targetWidths.js
```diff
@@ -0,0 +1,109 @@
...
+      r.push(arr[i]);
+    }
+  }
+
+  return r;
+}
+
+import CONSTANTS from "./constants";
+
+const _screen = {
```
jayeb: Is this constant even used?

README.md
frederickfogerty: Whoops, I was planning to do this but didn't get around to it. I might pull this out and put it in a separate issue

src/constants.js
frederickfogerty: I realise now this isn't used anymore anyway so I'll just remove it.

src/react-imgix.js
frederickfogerty: This is an extremely common pattern which is used everywhere in React. It is usually defined by webpack plugin or webpack itself.

src/react-imgix.js
frederickfogerty: Updated

src/react-imgix.js
frederickfogerty: My aim with `_srcSetWithFallback` was to make clear what the added `src` was for. Would this work?

```js
        const addFallbackSrc = srcSet => srcSet.concat(src);
        const _srcSet = targetWidths.map(buildSrcSetPair);
        const _srcSetWithFallback = addFallbackSrc(_srcSet);
        srcSet = _srcSetWithFallback.join(', ');
```

or even

```js
        const addFallbackSrc = srcSet => srcSet.concat(src);
        srcSet = addFallbackSrc(targetWidths.map(buildSrcSetPair)).join(', ');
```

src/targetWidths.js
frederickfogerty: No, I forgot to remove it when I did the update.

src/targetWidths.js
frederickfogerty: Thank you. That sounds like a great idea.

frederickfogerty:
Thanks for the review @jayeb. You're dead right with the constants - I've removed them.

README.md
frederickfogerty: Removed from PR and added to #161

frederickfogerty:
Okay @jayeb I think I've covered everything discussed here - are you able to give this a quick review?

PR #162, opened 2018-07-30
feat: reduce props API surface area

frederickfogerty:
BREAKING change: faces is no longer set by default.

## Description

This PR consolidates some imgix props into the `imgixParams` object, which makes the props API smaller and easier to understand. There should be no issues with excessive re-rendering as the object is checked shallowly to see if the object's values have changed.

In the past this library has set the `faces` imgix parameter by default. The new behaviour is to not set this parameter so that it should be less confusing for new-comers.

Also, I'm looking for good taglines/features about this library, so if you, reader, have any ideas, please let me know.

Ref #144

### New Feature

- [x] Update the readme

## Steps to Test

Review changes to unit tests.


src/react-imgix.js
```diff
@@ -116,16 +124,34 @@ export default class ReactImgix extends Component {
     };
   };
 
+  imgixParams = () => {
+    const params = {
+      ...defaultImgixParams,
+      ...this.props.imgixParams
+    };
+
+    const { faces, entropy } = params;
```
jayeb: I feel a little bit uneasy about encouraging users to put "non-standard" parameters into the `imgixParams` object. I have no problem putting "helper" options into the parent `this.props` object, but if we're going to explicitly call this a map of "imgix params" then I feel like we should require users to pass in `crop: 'faces'` instead of `faces: true`.

src/react-imgix.js
```diff
@@ -116,16 +124,34 @@ export default class ReactImgix extends Component {
...
+      fit
+    };
+  };
+
   render() {
     const {
-      auto,
       children,
       component,
-      customParams,
```
jayeb: `imgixParams` should be added to `this.props` here, no?

src/react-imgix.js
```diff
@@ -252,3 +279,54 @@ export default class ReactImgix extends Component {
...
+      ...props.imgixParams
+    };
+    const propsWithOutDeprecated = {
+      ...props,
+      imgixParams
+    };
+    DEPRECATED_PROPS.forEach(deprecatedProp => {
+      warning(
+        !(deprecatedProp in props),
+        `The prop '${deprecatedProp}' has been deprecated and will be removed in v9. Please update the usage to <Imgix imgixParams={{${deprecatedProp}: value}} />`
```
jayeb: Following my above comment about not encouraging users to pass in `{ imgixProps: { entropy: true } }`, this warning will need to be updated.

src/react-imgix.js
```diff
@@ -16,33 +18,33 @@ const buildKey = idx => `react-imgix-${idx}`;
...
     disableSrcSet: PropTypes.bool,
     onMounted: PropTypes.func,
     sizes: PropTypes.string,
     src: PropTypes.string.isRequired,
     type: PropTypes.oneOf(validTypes),
     width: PropTypes.number,
     height: PropTypes.number,
-    disableLibraryParam: PropTypes.bool
+    defaultHeight: PropTypes.number,
+    defaultWidth: PropTypes.number,
```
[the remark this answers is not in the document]
frederickfogerty: My bad, they were removed.

src/react-imgix.js
```diff
@@ -16,33 +18,33 @@ const buildKey = idx => `react-imgix-${idx}`;
 
 const validTypes = ["img", "picture", "source"];
 
-export default class ReactImgix extends Component {
+const defaultImgixParams = {
+  auto: ["format"],
+  fit: "crop"
```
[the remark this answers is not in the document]
frederickfogerty: As discussed offline, `fit=crop` works intuitively for developers, but it needs to be called out in the README somewhere.

src/react-imgix.js
```diff
@@ -116,16 +124,34 @@ export default class ReactImgix extends Component {
...
+
+    const { faces, entropy } = params;
+
+    let crop = false;
+    if (faces) crop = "faces";
+    if (entropy) crop = "entropy";
+    if (params.crop) crop = params.crop;
+
+    let fit = false;
+    if (entropy || faces) fit = "crop";
```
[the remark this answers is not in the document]
frederickfogerty: This sounds like a great idea.

src/react-imgix.js
frederickfogerty: I had actually forgotten that `faces` wasn't a top level imgix parameter. I agree with you, I feel uneasy about it as well. I'll remove `faces` in favour of using `crop: 'faces'` directly.

src/react-imgix.js
frederickfogerty: I don't think so -- this is not setting the prop types for the component, this is destructuring props into local variables to make it easier to use in the `render` method. `imgixParams` is not used anywhere in render, only in `ReactImgix.buildSrcs()`

frederickfogerty:
Hey @jayeb, I think I've addressed all the issues discussed - could you please give this another review?

PR #201, opened 2018-10-10
feat: add aspect ratio support by calculating client-side

### LINKED ISSUE #161: Ability to set aspectRatio
**Before you submit:**

- [x] Please search through the existing issues (both open AND closed) to see if your feature has already been discussed. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please take a moment to find out whether your idea fits with the scope and aims of the project

**Is your feature request related to a problem? Please describe.**
It's currently impossible to maintain the aspect ratio correctly when using `srcSet` and `sizes`.
A lot of the other Imgix libraries support the ability to set the aspectRatio or handle aspectRatio in their calculations.

An `aspectRatio` prop and for srcSets to respect the supplied aspect ratio.

frederickfogerty:
## Description

This PR allows the developer to specify a desired aspect ratio (AR) when creating an image. 

This PR is based off the awesome work from @maxkolyanov!

This fixes #161

### API

The developer specifies the AR along with normal imgix params, such as `fit`.

e.g.
```jsx
<Imgix
  src={src}
  sizes="(max-width: 30em) 100vw, (max-width: 50em) 50vw"
  imgixParams={{ ar: "2:1" }}
/>
```

The AR has to be specified in the form `number:number`. When an invalid AR is detected, no height values will be generated, and a warning is thrown.

### TODO

- [N/A] Audit tests
- [x] Round `h` up to comply with API spec 
- [x] Add Max as contributor

### New Feature

- [x] Update the readme

## Steps to Test

Review changes to unit tests.


src/react-imgix.js
```diff
@@ -47,6 +48,26 @@ const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
   src: PropTypes.string.isRequired
 };
 
+/**
+ * Parse an aspect ratio in the format w:h to a decimal. If false is returned, the aspect ratio is in the wrong format.
+ */
+function parseAspectRatio(aspectRatio) {
+  if (typeof aspectRatio == null) {
```
maxkolyanov: This is potentially redundant.
It would be sufficient to make sure that `aspectRatio` is a string.
So check below would be sufficient on its own.

src/react-imgix.js
```diff
@@ -80,15 +102,35 @@ function buildSrc({
...
     } else {
+      let showARWarning = false;
       const buildSrcSetPair = targetWidth => {
-        const url = constructUrl(rawSrc, {
+        let urlParams = {
           ...srcOptions,
           width: targetWidth
-        });
+        };
+        const aspectRatioDecimal = aspectRatio && parseAspectRatio(aspectRatio);
```
maxkolyanov: Following on comments on type checks in `parseAspectRatio`;
It seems to be sufficient to just pass `aspectRatio` into `parseAspectRatio` and throw warning if `false` was returned.

src/react-imgix.js
```diff
@@ -80,15 +102,35 @@ function buildSrc({
...
+        if (aspectRatio != null && aspectRatioDecimal === false) {
```
maxkolyanov: See above. 
If all type checks just happen inside `parseAspectRatio`, `aspectRatio != null` can be removed.

src/react-imgix.js
```diff
@@ -38,6 +38,7 @@ const COMMON_PROP_TYPES = {
 
 const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
   ...COMMON_PROP_TYPES,
+  aspectRatio: PropTypes.number,
```
maxkolyanov: It seems like `aspectRatio` is a string

src/react-imgix.js
```diff
@@ -47,6 +48,26 @@ const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
...
+    return false;
+  }
+  const isValidFormat = str => !/^\d(\.\d+)?:\d(\.\d+)?$/.test(str);
+  if (isValidFormat(aspectRatio)) {
+    return false;
+  }
+
+  const [width, height] = aspectRatio.split(":");
+
+  return width / height;
```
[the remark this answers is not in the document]
frederickfogerty: Are you saying this should be `return parseFloat(width) / parseFloat(height)`? I'm not quite understanding this comment

src/react-imgix.js
```diff
@@ -47,6 +48,26 @@ const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
...
+ */
+function parseAspectRatio(aspectRatio) {
+  if (typeof aspectRatio == null) {
+    return false;
+  }
+  if (typeof aspectRatio !== "string") {
+    return false;
+  }
+  const isValidFormat = str => !/^\d(\.\d+)?:\d(\.\d+)?$/.test(str);
+  if (isValidFormat(aspectRatio)) {
```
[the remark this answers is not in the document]
frederickfogerty: Good catch 🙌

src/react-imgix.js
frederickfogerty: The reason for having `aspectRatio != null` here is so that a warning is only thrown when the user has tried to pass something. e.g. it wouldn't throw a warning when no prop is passed, but it would when the user passes "".

src/react-imgix.js
frederickfogerty: This was fixed in a commit I didn't push 😬

src/react-imgix.js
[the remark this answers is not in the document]
maxkolyanov: That is exactly what I meant! 
Sorry, ironically I phrased it weirdly as well.

src/react-imgix.js
```diff
@@ -47,6 +47,23 @@ const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
...
 };
 
+/**
+ * Parse an aspect ratio in the format w:h to a decimal. If false is returned, the aspect ratio is in the wrong format.
+ */
+function parseAspectRatio(aspectRatio) {
+  if (typeof aspectRatio !== "string") {
+    return false;
+  }
+  const isValidFormat = str => /^\d(\.\d+)?:\d(\.\d+)?$/.test(str);
```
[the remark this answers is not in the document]
frederickfogerty: Good catch!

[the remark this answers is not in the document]
jayeb: Can we also update the unit tests to make sure this is covered?

frederickfogerty:
@maxkolyanov @jayeb I think I've covered everything that was brought up. Can I please get another quick review?

src/react-imgix.js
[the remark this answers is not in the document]
frederickfogerty: Yep, all done!

src/react-imgix.js
maxkolyanov: Good call! Totally makes sense to check if the prop is there, before showing a warning.
One questions I guess is wouldn't it be equal to `undefined` instead of `null`, if the prop is absent/not passed?

frederickfogerty: Definitely! This is the reason for using `!=` rather than `!==`, as this covers both `undefined` and `null`.

i.e.
```js
undefined != null // false
undefined !== null // true
```

maxkolyanov: Totally! Great work on PR, LGTM 👍

frederickfogerty: Thanks @maxkolyanov! 🙌 thanks again for your work for the base for this PR.

PR #224, opened 2018-11-17
feat: use exponential increase for srcset widths

frederickfogerty:
## Description

This PR updates the srcset width-generation logic. The widths now grow exponentially, rather than by a fixed amount. This means that the dimensions of a  rendered image are at most 8% from the "pixel-perfect" dimensions. 

Fixes #224

## Steps to test

Ensure that tests still pass.


jayeb:
Ah, I saw something about tests in the description and assumed there would be tests coming to cover the new behavior. Could we get some?

frederickfogerty:
Similar to the comment over on imgix.js (https://github.com/imgix/imgix.js/pull/130), not sure what tests I can add which don't over-specify on the implementation.

frederickfogerty:
Hey @jayeb I've added some more tests as discussed offline - can you have a look?

PR #248, opened 2018-12-19
Adds a comma to README

adamraider:
The example code had a syntax error


maniator:
:+1: Amazing!

frederickfogerty:
Thanks for the PR @adamraider! So we don't have to mess around with Prettier issues, I'm going to fix this over in #250.

Issue #129, opened 2018-03-19
Support Variable Qualities for SrcSet


AdamPflug:
**type:** feature request

When working with SrcSets I find it's pretty important to specify lower encoding qualities for the high-dpi images, otherwise you end up with your largest file sizes being served to mobile devices (which are most sensitive to file size). This is because phones/tablets tend to both display images full width instead of in columns, and are more likely to have high DPI screens. High resolutions with high compression ratios still look significantly better at similar file sizes to lower resolution uncompressed images on these displays, so it's still worth using srcsets rather than just leaving them off.

In fact, this mirrors the Imgix's own recommendation: https://docs.imgix.com/tutorials/responsive-images-srcset-imgix (See "Use Variable Quality")

The trick is, I can't find a way to specify parameters per-dpi with this library. My first thought about what the API would look like would be something like this:

```jsx
<Imgix src={url} faces={false} fit="max" generateSrcSet={[{q: 95}, {q: 75}, {q: 50}]} /> 
```

where each element in the array is basically a `customParams` object, but specifically targeted at that srcset entry (so the first one is 1x, the second one is 2x, the third is 3x). This is super flexible (supports more than just changing the quality), and would potentially even unlock custom dprs like this:

```jsx
<Imgix src={url} generateSrcSet={[{q: 95, dpr: 1}, {q: 75, dpr: 1.5}, {q: 50, dpr: 2}]} /> 
```

Thoughts? Am I missing something?

Thanks!

- Adam

frederickfogerty:
Hey Adam!

Thanks for the great suggestion! Your API does seem like an intuitive one, and would be a welcome addition to this library. 

Unfortunately I can't commit time to this project at the moment, and as such I can't guarantee I'd be able to develop this. If you have some time and would be happy to contribute this to the repository, I'd happily take a PR for this!

Cheers, Fred

frederickfogerty:
Hey there @AdamPflug - I'm now working for Imgix now, so I have more time to dedicate to features like this. After speaking with the team, we think is a great idea and we'll look to support this (or something with the same effect) soon - I can't give an exact timeline unfortunately.

AdamPflug:
Sounds great, thanks!

zachafranz:
Hey @frederickfogerty was just wondering about the status/progress of this feature?

jayeb:
[...]

PR #302, opened 2019-02-27
feat: support variable qualities #129

zachafranz:
## Description
Adds support for supplying variable imgix parameters to each of the different dpr urls generated in the srcset.

Would revert back to using `genSrcSet` (instead of the currently used prop `disableSrcSet`) as the prop to enable/disable generate of the different dpr urls for srcset. `genSrcSet` could either be a boolean, true to generate variable width srcset and false to not generate variable width srcset, or an array to generate variable width srcset with the ability to add custom imgix props for individual srcs in the srcset.

Compared to the original [issue](https://github.com/imgix/react-imgix/issues/129) opened, this feature would not allow the user to specify additional and custom dpr values for the srcset.

The addition of the feature would deprecate `disableSrcSet` and replace it with `genSrcSet` as well as flip the logic of the prop.

No new tests were added.

### New Feature
- [ x ] If this is a big feature with breaking changes, consider [opening an issue][issues] to discuss first. This is completely up to you, but please keep in mind that your PR might not be accepted.
- [ x ] Run unit tests to ensure all existing tests are still passing
- [   ] Add new passing unit tests to cover the code introduced by your PR
- [ x ] Update the readme
- [ x ] Add some [steps](#steps-to-test) so we can test your cool new feature!
- [ x ] The PR title is in the [conventional commit](#conventional-commit-spec) format: e.g. `feat(<area>): added new way to do this cool thing #issue-number`
- [ x ] Add your info to the [contributors](#packagejson-contributors) array in package.json!

## Steps to Test
1. Use the following code in any preferred sandbox environment 
```js
import Imgix from "react-imgix";

<Imgix src="https://assets.imgix.net/examples/pione.jpg" genSrcSet={[{q: 95, dpr: 1}, {q: 75, dpr: 2}, {q: 50, dpr: 3}]} />;
```
2. Inspect the element to see variable qualities on each of the srcs in the srcset. Srcs with dpr 4 and 5 will have `q` of 95 as the `q` of the dpr 1 src acts as a default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,13 +388,25 @@ Force images to be a certain height.
 
 Force images to be a certain width.
 
-##### disableSrcSet :: bool, default = false
-
-Disable generation of variable width src sets to enable responsiveness.
-
 ##### disableLibraryParam :: bool
 
 By default this component adds a parameter to the generated url to help imgix with analytics and support for this library. This can be disabled by setting this prop to `true`.
+
+##### genSrcSet :: bool | array, default = true
+
+As a boolean, enable generation of variable width src sets to enable responsiveness. As an array, enable generation of variable width src sets and allows setting of different imgix params to each src in the src set.
+
+_For example_:
+
+```js
+import Imgix from "react-imgix";
+
+// As a boolean. Disables generation of variable width src sets.
+<Imgix src="https://assets.imgix.net/examples/pione.jpg" genSrcSet={false} />;
+
+// As an array.
+<Imgix src="https://assets.imgix.net/examples/pione.jpg" genSrcSet={[{q: 95, dpr: 1}, {q: 75, dpr: 2}, {q: 50, dpr: 3}]} />;
+```
 
 ##### htmlAttributes :: object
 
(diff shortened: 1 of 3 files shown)


sherwinski:
Hey @zachafranz thanks a lot for taking the time to build this out. Although you’re on the right track here, I don’t think we are going to take in these changes due to some implementation concerns that I will explain below.

First, the decision to deprecate `disableSrcSet` and to flip the logic with `genSrcSet` is something we’d rather avoid. In most cases users will want to generate srcset with non-fixed sizes, so we would want to keep that consistent. That means that users who specifically want a fixed size srcset generated should have to expressly do so. Also the method of passing in variable qualities seems a bit messy and may be unintuitive to users with all the nested objects/key-value pairs that are going on.
After discussing it with @jayeb we think we know how we’d like to add this feature in way that adheres to our interface design so far. Users will basically pass in a `true` value to create a fixed set of `q` values matched against each `dpr` set. Which will also respect any overwriting `q` values a user might supply. 

And since you’ve already got the ball rolling on this, we’re going to add it into our roadmap and hopefully have something ready within the next few weeks. I’d love to add you as a reviewer when we get to that point since you’ve already weighed in on this quite a bit. Feel free to comment back with any questions you have!

zachafranz:
@sherwinski Thank you for taking the time to look at this. The concerns you bring up are certainly understandable and I felt some of them when I created this PR in the first place.

Thank you for putting this on your roadmap. I look forward to seeing the implementation the team arrives at and getting this feature available. Would be happy to review when something is ready 👍.

PR #322, opened 2019-03-22
feat: append variable q parameters per dpr when rendering a fixed-size image

### LINKED ISSUE #129: Support Variable Qualities for SrcSet

sherwinski:
## Description

This addresses the enhancement request made for supporting variable qualities and resolves #129. Special thanks to @zachafranz for getting the ball rolling on this feature.

This PR adds new behavior that will automatically append a `q` parameter mapped to each `dpr` parameter when generating a fixed-size srcset. This technique is commonly used to compensate for the increased filesize of high-DPR images. Since high-DPR images are displayed at a higher pixel density on devices, image quality can be lowered to reduce overall filesize without sacrificing perceived visual quality. For more information and examples of this technique in action, see [this blog post](https://blog.imgix.com/2016/03/30/dpr-quality).
This behavior will respect any overriding `q` value passed in via the `imgixParams` property. Additionally, it can be disabled altogether with the new boolean property `disableQualityByDPR`.

Based on a [visual demo](https://codepen.io/imgix/full/WmLYyY) comparing different quality/device pixel ratio combinations, the following map of `q` to `dpr` pairs will be generated:

`dpr` | `q`
--- |---
1 | 75 
2 | 50 
3 | 35 
4 | 23 
5 | 20 

This behavior specifically occurs when a fixed-size image is rendered, or anytime that a `width` and/or `height` are explicitly passed in to the component, for example:

```js
<Imgix src="https://domain.imgix.net/image.jpg" width={100} />
```
will generate the following srcset:

```html
https://domain.imgix.net/image.jpg?q=75&w=100&dpr=1 1x,
https://domain.imgix.net/image.jpg?q=50&w=100&dpr=2 2x, 
https://domain.imgix.net/image.jpg?q=35&w=100&dpr=3 3x, 
https://domain.imgix.net/image.jpg?q=23&w=100&dpr=4 4x, 
https://domain.imgix.net/image.jpg?q=20&w=100&dpr=5 5x
```

- [x] Update the readme

## Steps to Test

- A [visual demo](https://codepen.io/imgix/pen/WmLYyY) comparing different q + dpr pairs
- Tests are included in `test/unit/react-imgix.test.jsx`


zachafranz:
@sherwinski this looks great and seems like a step in the right direction. Is it possible to give the user control over the different quality/device pixel ratio combinations when using `<Imgix/>`?

sherwinski:
Hey @zachafranz to answer your question, no we didn't include a way for users to pass in different quality/dpr combinations. A lot of that decision ties back to the discussion we had in #302 about it not really being possible without being messy at the same time. We did a lot of [research and testing](https://codepen.io/imgix/pen/WmLYyY) before building this feature and basically came up with the combination set described in the description above. We think that for a majority of cases, these should provide the lowest quality values without a noticeable loss in image quality.

But maybe you have a use case that isn't covered by these changes? It would be great to hear what you're looking for more specifically.

README.md
```diff
@@ -190,6 +190,21 @@ import Imgix from "react-imgix";
 />;
 ```
 
+As of version 8.6.0, fixed image rendering will automatically append a `q` parameter matched against each `dpr` parameter when generating a fixed-size srcset. This behavior will respect any overriding `q` value passed in via `imgixParams` and can be disabled altogether with the new boolean property `disableQualityByDPR`.
```
jayeb: Can we add a sentence here explaining why we do this? I'm worried that this probably won't be obvious to most folks and they'll be left in confusion after reading this paragraph. Something like this, maybe?

> This technique is commonly used to compensate for the increased filesize of high-DPR images. Since high-DPR images are displayed a higher pixel density on devices, image quality can be lowered to reduce overall filesize without sacrificing perceived visual quality. For more information and examples of this technique in action, see [this blog post](https://blog.imgix.com/2016/03/30/dpr-quality).

README.md
```diff
@@ -190,6 +190,21 @@ import Imgix from "react-imgix";
...
+
+```js
+<Imgix src="https://domain.imgix.net/image.jpg" width={100} />
+```
+
+will generate the following srcset:
+
+```html
+https://domain.imgix.net/image.jpg?q=50&w=100&dpr=2 2x,
```
jayeb: Is there a reason you left the `dpr=1` image out of this example?

src/react-imgix.jsx
```diff
@@ -94,10 +96,26 @@ function buildSrc({
...
+        ...(disableQualityByDPR || { q: CONSTANTS.q_dpr4 }),
+        ...srcOptions,
+        dpr: 4
+      });
+      const dpr5 = constructUrl(rawSrc, {
+        ...(disableQualityByDPR || { q: CONSTANTS.q_dpr5 }),
+        ...srcOptions,
+        dpr: 5
+      });
       srcSet = `${dpr2} 2x, ${dpr3} 3x, ${dpr4} 4x, ${dpr5} 5x`;
```
jayeb: Wait. We don't include a `1x` URL in the `srcset` at all here? That seems wrong. Can we remedy that?

sherwinski: For some reason, I assumed that's what the fallback `src` was for but now realize that's not the case.

README.md
sherwinski: Totally, that's a great point.

src/react-imgix.jsx
jayeb: It's possible that some browsers would fall back to the `src` if they didn't find a DPR item in the `srcset` that they like, but we definitely shouldn't rely on that behavior.

PR #371, opened 2019-05-28
Fix docs on how to add a class to an img element

osnysantos:
Looks like `htmlAttributes={{ className: 'foo' }}` is not working as expected. I tried `htmlAttributes={{ class: 'foo' }}` and worked fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ By default this component adds a parameter to the generated url to help imgix wi
 
 ##### htmlAttributes :: object
 
-Any other attributes to add to the html node (example: `alt`, `data-*`, `className`).
+Any other attributes to add to the html node (example: `alt`, `data-*`, `class`).
 
 ##### onMounted :: func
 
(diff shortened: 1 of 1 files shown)


frederickfogerty:
Hey @osnysantos! The real issue here is that we override any `className` that's present in htmlAttributes with `this.props.className`. The implementation can be seen [here](https://github.com/imgix/react-imgix/blob/master/src/react-imgix.jsx#L233). I'm not sure if this is the best behavior, so if you're willing to submit a PR that merges `props.className` and `props.htmlAttributes.className`, I'd be happy to accept that!

The workaround for this at the moment is to use `className` directly on the imgix component! So, this `<Imgix className="my-class" />` rather than `<Imgix htmlAttributes={{ className: "my-class" }} />`

I'm not going to be able to accept this change because passing `class` as a prop to a React component will throw a warning (or at least, it used to do this).

PR #384, opened 2019-06-08
docs(cleanup): remove fit-to-size-of-containter example

sherwinski:
Removing this page as it is no longer needed.


sherwinski:
Yep exactly

Issue #387, opened 2019-06-14
Using `<Background />` tanks my CPU


drydenwilliams:
**Describe the bug**
I have a media gallery which loads set background images using the `<Background />` component and as soon as I started using it, my CPU load spike massively. 

![Screenshot 2019-06-14 at 13 05 53](https://user-images.githubusercontent.com/4403089/59508308-606cc580-8ea5-11e9-83ab-926bdcede29b.png)

Different between not using the `<Background />` component in the first half of the graph. And using it in the second half. 

<img width="557" alt="Screenshot 2019-06-14 at 13 11 03" src="https://user-images.githubusercontent.com/4403089/59508474-fdc7f980-8ea5-11e9-9c3f-2849fbe339d7.png">

**To Reproduce**
Add a list of images using the `<Background />` component and see the spike in CPU.

```
<Background
  src={imageSrc}
  imgixParams={{
    crop: 'entropy',
    h: 200,
    w: 200
  }}
  className="blog-title"
>
 Title
</Background>
```
**Information:**

- react-imgix version: [e.g. v8.6.1]
- browser version: https://whatsmybrowser.org/b/RAF3DEX(https://www.whatsmybrowser.org/

ericdeansanchez:
Hey @drydenwilliams, I'm currently working on solving this issue. The `<Background/>` component is constantly re-rendering. The fix I'm working on aims to ensure minimal re-renders.

Issue #328, opened 2019-03-30
Server-Side Rendering srcSet generation performance bottleneck


tstirrat15:
When profiling our isomorphic React app, one thing that we noticed was that srcSet generation was taking up a whole lot of CPU time - somewhere between a third and a half of the total cpu render time on a page that had 17 images.

Here's a [`node-clinic` flamegraph](https://upload.clinicjs.org/public/b504341d486f38f96cea706cce88906329b7fdc576d8d105dacd36f1e9cc593a/9042.clinic-flame.html#selectedNode=3007&zoomedNode=&exclude=83c0&merged=true) showing the problem.

We were able to work around the problem by disabling srcset generation on the backend, especially because the images that were being sent from the backend were LQIPs, and it didn't make sense to have a full srcset for those, but the amount of CPU time consumed by generating srcSets was surprising to me.

Has this already been optimized, and it's just an inherently expensive operation? Or are there gains to be made in optimizing the srcSet generation code?

sherwinski:
Hi @tstirrat15 thanks for bringing this up. As far as I know, srcset generation has not been exhaustively optimized so perhaps there are gains to be made there. I won't be able to dedicate time to look into this just yet, but would happily review a PR or discuss anything you come across if you have the time and want to look into it.

sherwinski:
hey @tstirrat15,

There have been numerous performance changes made to URL building (all credit goes to @baldurh) since this issue was first opened. I would invite you, or anyone else who's curious, to test these out in the latest release. I'm going to go ahead and close this issue now, but feel free to comment back with any questions you may have.

Also relates to #406, #414, #418

tstirrat15:
Thank you!

PR #395, opened 2019-06-28
feat: optimize url construction

baldurh:
## Description

This PR tries to address the issues presented in the issues #328 and #387 by optimizing the code used to construct urls.

- Removes usage of `jsuri` from `constructUrlFromParams` and replaces it with custom code
- Optimizes how `buildSrc` constructs `srcSet` by reusing the constructed url.

I’m aware that introducing custom code for managing query parameters is a potential risk.

Related issue: #328, #387

**Flamegraph before**
image

**Flamegraph after**
image

**Flamegraph final** (including changes to fixed size url calculations)
image


src/constructUrl.js
```diff
@@ -106,12 +105,18 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+  const oldParams = query ? query.split("&").map(x => x.split("=")) : [];
+  const allParams = {
+    ...oldParams.reduce(
+      (obj, [key, val]) => Object.assign(obj, { [key]: val }),
+      {}
+    ),
+    ...params
+  };
+  return `${url}?${Object.entries(allParams)
+    .map(([a, b]) => `${a}=${encodeURIComponent(b)}`)
```
jayeb: This would double-encode any parameters that came in pre-encoded in `src`, right? Perhaps those parameters should be naively decoded during the map step on L109.

src/react-imgix.jsx
```diff
@@ -9,6 +9,7 @@ import constructUrl from "./constructUrl";
 import { deprecatePropsHOC, ShouldComponentUpdateHOC } from "./HOCs";
 
 import { warning, shallowEqual, compose, config, CONSTANTS } from "./common";
+import { isGenericTypeAnnotation } from "@babel/types";
```
jayeb: How does this work, exactly? Importing this type annotation affects the output file, even though the imported object is never referenced?

baldurh: Whoops! VS code must have added this automatically when I’ve mistakenly used some suggested option. I’ll remove this and rebase. This slipped passed me. Nice catch ✌️

src/constructUrl.js
baldurh: Yes. Good point. I didn’t think of that.

baldurh:
@jayeb I’m happy to help 🙂
I’ve addressed your comments. Please take a look

src/react-imgix.jsx
```diff
@@ -124,25 +124,28 @@ function buildSrc({
...
+      const { width, height, ...urlParams } = srcOptions;
+      const constructedUrl = constructUrl(rawSrc, urlParams);
+
+      const aspectRatioDecimal = parseAspectRatio(aspectRatio);
+      if (aspectRatio != null && aspectRatioDecimal === false) {
+        // false indicates invalid
+        showARWarning = true;
+      }
+
       const buildSrcSetPair = targetWidth => {
```
frederickfogerty: Optional: this function could be moved out into the module scope, and the variables used passed in as parameters.

src/react-imgix.jsx
```diff
@@ -124,25 +124,28 @@ function buildSrc({
...
+          url = `${url}&h=${height}`;
         }
-        if (
-          !srcOptions.height &&
+        else if (
           aspectRatioDecimal &&
           aspectRatioDecimal > 0
         ) {
-          urlParams.height = Math.ceil(targetWidth / aspectRatioDecimal);
+          url = `${url}&h=${Math.ceil(targetWidth / aspectRatioDecimal)}`;
```
frederickfogerty: Optional: maybe there's more performance to be gained from doing this concatenation (along with L139) in one go with L148? (the template string in the return statement). Something like 

```js
let h;

// ...

h = Math.ceil(targetWidth / aspectRatioDecimal);

// ...

return `${url}&h={h}&w=${targetWidth} ${targetWidth}w`; 
```

src/constructUrl.js
```diff
@@ -5,7 +5,6 @@ Licensed under the Apache License 2.0, seen https://github.com/coursera/react-im
 Minor syntax modifications have been made
 */
 
-var Uri = require("jsuri");
```
frederickfogerty: Now that jsuri is only used in the tests, this can moved to a devDep 👍

src/react-imgix.jsx
frederickfogerty: If this is done, I'm also curious as to what effect memoization could have on the performance of this function. My guess is that it wouldn't improve performance... but miracles can happen.

src/react-imgix.jsx
baldurh: In that case we might get `&h=&w=...`. Is that acceptable?

baldurh: Never mind. I changed the behaviour a little bit 🙂

baldurh:
@frederickfogerty Thanks for your suggestions 🙂
I made some changes. Mainly, I moved `buildSrcSetPair` out of `buildSrc` like you suggested and I changed the function so it does as little work as possible for each srcSet pair. It basically returns different build functions based on whether there is a fixed height, aspect ratio or no height given. This way we are pretty much "memoizing" since the only work being done for each srcSet pair is just appending the width.

baldurh:
Basic time measurements (on my laptop)
Running `buildSrc` 10,000 times
**Current version**: 3698.878ms
**This branch**: 876.097ms

frederickfogerty:
@baldurh Prettier doesn't seem happy with the code formatting here - this can be fixed by running `npm run pretty` and committing the changes. Usually Prettier would auto-fix this for us but it doesn't work with outsider PRs 😬.

If you have any more issues with Prettier, tag me and I'll jump in to help.

baldurh:
@frederickfogerty yeah I noticed that and I already took care to run it. Now when I run it there are no changes locally. So I guess the configuration is a bit off maybe? Also I can't see what the problem is when I click the thing so I can't fix it manually either 🙈

frederickfogerty:
@baldurh I ran `npm i && npm run pretty` and there was a change... I just can't quite work out what it is. Potentially a line ending, as it doesn't look like whitespace.

Here's a `git diff`:

image

frederickfogerty:
@baldurh I went and edited the file directly, and Prettier seems happy now. Just waiting on a 👍 from @jayeb and we should be good to go!

baldurh:
@frederickfogerty that's so weird. I don't know why it didn't work for me. Anyway, thanks for your help 👌🏻

frederickfogerty:
@baldurh Usually the culprit is an out-of-date prettier installation (global or local) 👍. But of course there are other reasons as well. Glad I could help out!

baldurh:
@jayeb Any chance you could take another look? We have a pending PR in our project we don’t want to merge until this is added 😬

sherwinski:
Hey @baldurh 👋 it looks like @jayeb will be unavailable to look at this until Friday if you're able to wait until then.

baldurh:
@sherwinski sure no problem 😊

baldurh:
> This is so clean now, I love it. Thanks for all the hard work @baldurh.

My pleasure. This is a fun project 🙂

sherwinski:
@baldurh these changes are now available in [v8.6.2](https://github.com/imgix/react-imgix/releases/tag/v8.6.2) 🎉 
Thanks again for all your help!

PR #404, opened 2019-07-09
fix: render <Source> element as a fluid image by default

### LINKED ISSUE #243: Use sizes on <Source> to generate srcSet
**Before you submit:**

- [x] Please search through the existing issues (both open AND closed) to see if your issue has been discussed before. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please ensure the problem has been isolated and reduced. This link explains more: http://css-tricks.com/6263-reduced-test-cases/

**Describe the bug**

It looks like the `<Source>` component ignores the `sizes=` attribute and just generates a `srcSet` with different resolutions. I think it's [this line](https://github.com/imgix/react-imgix/blob/master/src/react-imgix.jsx#L96):
```
if (fixedSize || type === "source") {
```
But I _think_ `<source>` should be able to support fluid width images the same as ``? Is there a reason that `<Source>` components are grouped in with fixed size images?

**To Reproduce**
```
<Picture>
  <Source
    src={desktopSrc}
    sizes="100vw"
    htmlAttributes={{ media: "(min-width: 768px)" }}
  />
  <Source
    src={mobileSrc}
    sizes="100vw"
    htmlAttributes={{ media: "(min-width: 320px)" }}
  />
  <Imgix src={mobileSrc} sizes="100vw" />
</Picture>
```

**Expected behaviour**

I'd expect a `<Source>` to generate its `srcSet` based on `sizes=` similar to the `<Imgix>` component - otherwise art-directed images end up much larger than they need to be.

sherwinski:
Resolves #243 

Previously, `<Source>` elements were generated with a `srcSet` attribute defined in DPR mode by default, regardless of whether a width and/or height prop were passed in.

This PR modifies the `srcSet` generation in `<Source>` elements to mimic the same behavior as responsive images generated by the `<Imgix>` component. Now the `srcSet` will be generated with width descriptors by default unless either or both dimensions are defined.

In addition, this PR addresses two bugs:
- the <Source> component was appending the fallback `src` to the `srcSet` attribute
- the largest responsive width generated in a given `srcSet` was greater than the enforced `max_width`

### DIFF
--- a/src/react-imgix.jsx
+++ b/src/react-imgix.jsx
@@ -131,7 +131,7 @@ function buildSrc({
   if (disableSrcSet) {
     srcSet = src;
   } else {
-    if (fixedSize || type === "source") {
+    if (fixedSize) {
       const { q: qOption, ...urlParams } = srcOptions;
 
       // Get the q from the raw src.
@@ -386,7 +386,7 @@ class SourceImpl extends Component {
     if (disableSrcSet) {
       childProps[attributeConfig.srcSet] = src;
     } else {
-      childProps[attributeConfig.srcSet] = `${src}, ${srcSet}`;
+      childProps[attributeConfig.srcSet] = `${srcSet}`;
     }
     // for now we'll take media from htmlAttributes which isn't ideal because
     //   a) this isn't an <img>
--- a/src/targetWidths.js
+++ b/src/targetWidths.js
@@ -10,7 +10,7 @@ function targetWidths() {
     resolutions.push(ensureEven(prev));
     prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
   }
-  
+
   resolutions.push(MAX_SIZE);
   return resolutions;
 }
--- a/src/targetWidths.js
+++ b/src/targetWidths.js
@@ -6,12 +6,12 @@ function targetWidths() {
 
   const ensureEven = n => 2 * Math.round(n / 2);
 
-  resolutions.push(prev);
   while (prev <= MAX_SIZE) {
-    prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
     resolutions.push(ensureEven(prev));
+    prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
   }
-
+  
+  resolutions.push(MAX_SIZE);
   return resolutions;
 }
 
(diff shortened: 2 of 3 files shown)


test/unit/react-imgix.test.jsx
```diff
@@ -237,19 +237,65 @@ describe("When in <source> mode", () => {
...
+        .split(", ")
+        .map(srcset => srcset.split(" ")[1])
+        .map(width => width.slice(0, -1))
+        .map(Number.parseFloat);
+
+      const min = Math.min(...srcsetWidths);
+      const max = Math.max(...srcsetWidths);
+
+      expect(min).not.toBeLessThan(100);
+      expect(min).not.toBeGreaterThan(8192);
```
sherwinski: @jayeb I assume this should be `max`? It appeared this same way elsewhere for a different test. Also, is this still the maximum width we want to enforce?

sherwinski:
@dbow Feel free to take a look as well since you originally opened the issue

test/unit/react-imgix.test.jsx
jayeb: Yea, seems like it should be `max`. And yes, this is still the maximum width our painters will render so it makes sense to enforce it as the maximum image size in the srcset.

src/react-imgix.jsx
```diff
@@ -386,7 +386,7 @@ class SourceImpl extends Component {
     if (disableSrcSet) {
       childProps[attributeConfig.srcSet] = src;
     } else {
-      childProps[attributeConfig.srcSet] = `${src}, ${srcSet}`;
```
jayeb: Interesting that this was in here as a direct rebuke to the helpful comment above it 🤔.

test/unit/react-imgix.test.jsx
sherwinski: Ah interesting because the largest width generated is currently 8584, but I think I see how that slipped through.

src/targetWidths.js
```diff
@@ -6,12 +6,12 @@ function targetWidths() {
...
   const ensureEven = n => 2 * Math.round(n / 2);
 
-  resolutions.push(prev);
   while (prev <= MAX_SIZE) {
-    prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
     resolutions.push(ensureEven(prev));
+    prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
   }
 
+  resolutions.push(MAX_SIZE);
```
sherwinski: @jayeb how do we feel about this? It doesn't abide by the 16% increment but it does feel right to include the max render size in the srcset.

jayeb: 👍

PR #423, opened 2019-07-24
perf: remove object spreading

baldurh:
## Description

I replaced all instances of the object spread operator with `Object.assign`. The results are quite remarkable:

Benchmarks in ms. 100,000 iterations

|                                       | before   | after    | change  |
|---------------------------------------|----------|----------|---------|
| responsive                            | 1652.000 | 1104.215 | -33.16% |
| responsive with fixed height          | 1373.051 | 886.121  | -35.46% |
| responsive with aspect ratio          | 1862.155 | 1267.419 | -31.94% |
| fixed width                           | 1347.304 | 867.349  | -35.62% |
| fixed width with quality              | 1420.417 | 908.058  | -36.07% |
| fixed width with dpr quality disabled | 1305.567 | 872.474  | -33.17% |
| fixed width with aspect ratio         | 1326.964 | 861.080  | -35.11% |

[Flamegraph before](https://upload.clinicjs.org/public/6e0b776202007313e42dfa1e7f18d8ea5775a7420a8413c35b014aaa43fb29ae/43077.clinic-flame.html#selectedNode=182&zoomedNode=&exclude=83c0&merged=true)
[Flamegraph after](https://upload.clinicjs.org/public/c9cca6135cc09e317485e3ec8c4f5309bb8e07fddb9062ac9872c8c534e793a6/43314.clinic-flame.html#selectedNode=159&zoomedNode=&exclude=83c0&merged=true)

### DIFF
--- a/src/constructUrl.js
+++ b/src/constructUrl.js
@@ -145,11 +145,15 @@ function buildURLPublic(src, imgixParams = {}, options = {}) {
 
   const [rawSrc, params] = extractQueryParams(src);
 
-  return constructUrl(rawSrc, {
-    ...params,
-    ...imgixParams,
-    ...(disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` })
-  });
+  return constructUrl(
+    rawSrc,
+    Object.assign(
+      {},
+      params,
+      imgixParams,
+      disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` }
+    )
+  );
 }
 
 export default constructUrl;
--- a/src/react-imgix.jsx
+++ b/src/react-imgix.jsx
@@ -38,17 +38,20 @@ const COMMON_PROP_TYPES = {
   htmlAttributes: PropTypes.object
 };
 
-const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = {
-  ...COMMON_PROP_TYPES,
-  disableQualityByDPR: PropTypes.bool,
-  disableSrcSet: PropTypes.bool,
-  disableLibraryParam: PropTypes.bool,
-  imgixParams: PropTypes.object,
-  sizes: PropTypes.string,
-  width: PropTypes.number,
-  height: PropTypes.number,
-  src: PropTypes.string.isRequired
-};
+const SHARED_IMGIX_AND_SOURCE_PROP_TYPES = Object.assign(
+  {},
+  COMMON_PROP_TYPES,
+  {
+    disableQualityByDPR: PropTypes.bool,
+    disableSrcSet: PropTypes.bool,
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ npm-debug.log
 es
 lib
 node_modules
+.clinic
```
frederickfogerty: We'd rather not have non-package-related exclusions checked into source control. Directories and files can be ignored locally (and not checked into source control) by moving this line to `.git/info/exclude`. This command will do it for you: `echo .clinic >> .git/info/exclude`

frederickfogerty:
Thanks again @baldurh! LGTM once that `.gitignore` change is fixed 👍

.gitignore
baldurh: Roger. Fixed!

frederickfogerty:
Food for thought here - in Chrome 76+, spread is actually faster than Object.assign 🤷‍♀️

image
Source: https://jsperf.com/object-assign-vs-spreading

baldurh:
> Food for thought here - in Chrome 76+, spread is actually faster than Object.assign 🤷‍♀️

Wow that’s super interesting. We definitely have to take that into account. I ran the benchmarks on node 12 and `Object.assign is` still faster than object spreading. Chrome 76 uses V8 7.6 while Node 12 is still at 7.4 so I guess that’s the reason. So I guess the question is: Do we want to make this faster for the latest Chrome browsers or everything else, including all Node servers (since none of them are running V8 7.6 yet)? Node servers will also probably not be using the latest V8 for a while even if the latest version of Node has it.

frederickfogerty:
Pulling in @jayeb to answer this from imgix's perspective. From my perspective, it's a tough call as there's no clear answer. I think we should accept this change as it is such a significant improvement, but aim to remove this in about a year's time (this could be put in a comment).

jayeb:
I may be off-base here, but it seems to me that the performance of the spread operator in Chrome is mostly irrelevant here--if this code is running in a browser, it's almost *certainly* going to be the Babel-compiled version, right? And since the compiled version isn't using the spread operator directly, it's kind of a moot point. This PR looks like an obvious win to me, but maybe I'm missing something.

baldurh:
> I may be off-base here, but it seems to me that the performance of the spread operator in Chrome is mostly irrelevant--if this code is running in a browser, it's almost _certainly_ going to be the Babel-compiled version, right? And since the compiled version isn't using the spread operator directly, it's kind of a moot point. This PR looks like an obvious win to me, but maybe I'm missing something.

@jayeb Yes you’re absolutely right. I realised this also this morning but hadn’t gotten around to mention it 😄 So babel transpiles the spread operator to a function looping over the props of the object being spread. So that’s going to be slow in any environment 🙈

PR #462, opened 2019-10-03
refactor: use ar parameter instead of calculating aspect ratio

sherwinski:
This PR refactors how the `ReactImgix` component determines an image's aspect ratio, now using the [imgix aspect ratio parameter](https://blog.imgix.com/2019/07/17/aspect-ratio-parameter-makes-cropping-even-easier) `ar=` instead of appending a calculated height `h=` value based on the prop. This should yield overall better performance, code cleanliness, and bring the component more in line with the imgix API.

---------
**Demo**
Given the same component invocation, the URL constructed will be different but still render the same image.
```jsx
<Imgix
	src="http://domain.imgix.net/image.jpg"
	imgixParams={{ ar: "2:1" }}
/>
```

|        | Example srcset URL                                        |
|--------|-----------------------------------------------------------|
| Before | http://domain.imgix.net/image.jpg?...&h=50&w=100          |
| After  | http://domain.imgix.net/image.jpg?...&ar=2%3A1&w=100      |

### DIFF
--- a/src/react-imgix.jsx
+++ b/src/react-imgix.jsx
@@ -73,21 +73,6 @@ function parseAspectRatio(aspectRatio) {
 const buildSrcSetPairWithFixedHeight = (url, targetWidth, fixedHeight, _) =>
   url + "&h=" + fixedHeight + "&w=" + targetWidth + " " + targetWidth + "w";
 
-const buildSrcSetPairWithAspectRatio = (
-  url,
-  targetWidth,
-  _,
-  aspectRatioDecimal
-) =>
-  url +
-  "&h=" +
-  Math.ceil(targetWidth / aspectRatioDecimal) +
-  "&w=" +
-  targetWidth +
-  " " +
-  targetWidth +
-  "w";
-
 const buildSrcSetPairWithTargetWidth = (url, targetWidth, _1, _2) =>
   url + "&w=" + targetWidth + " " + targetWidth + "w";
 
@@ -111,7 +96,6 @@ function buildSrc({
   disableSrcSet,
   type,
   imgixParams,
-  aspectRatio,
   disableQualityByDPR
 }) {
   const fixedSize = width != null || height != null;
@@ -156,6 +140,7 @@ function buildSrc({
       const { width, w, height, ...urlParams } = srcOptions;
       const constructedUrl = constructUrl(rawSrc, urlParams);
 
+      const aspectRatio = imgixParams.ar;
       const aspectRatioDecimal = parseAspectRatio(aspectRatio);
       // false indicates invalid
       let showARWarning = aspectRatio != null && aspectRatioDecimal === false;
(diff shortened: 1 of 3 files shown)


test/unit/react-imgix.test.jsx
```diff
@@ -648,43 +648,39 @@ describe("When using the component", () => {
...
           ar: validAR,
-          arDecimal: validArDecimal
         })
       );
     });
 
     describe("invalid AR", () => {
       const testInvalidAR = ar => {
-        it(`height should not be set when an invalid aspectRatio (${ar}) is passed`, () => {
+        it(`an invalid ar prop (${ar}) will still generate an ar query parameter`, () => {
```
sherwinski: @jayeb Typically, passing an invalid `ar` value/format would result in a warning and then ignoring the rest of the aspect ratio operation. 
While I definitely like the extra layers set in place to protect against user error, I don't think we should treat any single imgix parameter differently than the others. For that reason, I think the implementation should still pass through the (incorrect) `ar` value, as is the case with other imgix libraries. 
However, I am still debating as to whether we should keep the warning or not. Thoughts?

frederickfogerty: 👍 for the warning (but I'm biased, I wrote it 😂)

test/unit/react-imgix.test.jsx
```diff
@@ -648,43 +648,39 @@ describe("When using the component", () => {
...
-        ["11.123:11.123", "1"]
+        ["1:1"],
+        ["1.1:1"],
+        ["1.12:1"],
+        ["1.123:1"],
+        ["1:1.1"],
+        ["1:1.12"],
+        ["1.1:1.1"],
+        ["1.123:1.123"],
+        ["11.123:11.123"]
```
frederickfogerty: These seem extraneous now as we don't do any pre-processing.

src/react-imgix.jsx
```diff
@@ -156,22 +140,22 @@ function buildSrc({
       const { width, w, height, ...urlParams } = srcOptions;
       const constructedUrl = constructUrl(rawSrc, urlParams);
 
+      const aspectRatio = imgixParams.ar;
       const aspectRatioDecimal = parseAspectRatio(aspectRatio);
       // false indicates invalid
       let showARWarning = aspectRatio != null && aspectRatioDecimal === false;
```
frederickfogerty: Now that we don't need the parsing behaviour of this function, we should probably make the following changes:

- remove `parseAspectRatio` and instead have a validator function
- remove `aspectRatioDecimal` as it's no longer needed
- only run the validator function in development

sherwinski: > only run the validator function in development

Mind elaborating a bit more on this, I think I'm missing something here. Thought it might be related to `config.warnings.invalidARFormat` but not sure.

frederickfogerty:
```js
if (process.env.NODE_ENV !== 'production' && config.warnings.invalidARFormat && !arIsValid(aspectRatio)) {
  console.warn(...);
}
```

frederickfogerty: Or potentially whatever the new version of that `process.env.NODE_ENV` pattern is. I think that's still relevant though.

PR #471, opened 2019-10-16
build(deps): add babel-plugin-transform-es2017-object-entries

### LINKED ISSUE #469: Object.values is not a function
`Object.values` is not supported by older browsers. Either make sure it is transpiled by babel or just replace this line https://github.com/imgix/react-imgix/blob/master/src/constants.js#L9 with something else.

P.S. This was added by me a while back. Sorry 😅

sherwinski:
Fixes #469 

Adds the [transform-es2017-object-entries](https://www.npmjs.com/package/babel-plugin-transform-es2017-object-entries) babel plugin for transpiling `Object.values`


package.json
```diff
@@ -134,6 +135,7 @@
   },
   "browserslist": [
     "ie 11",
+    "ie 10",
```
sherwinski: In an attempt to test this, I tried to get browserstack to trip up an error by testing on `ie 10` - which apparently won't work for whatever reason.
In the [original issue](https://github.com/imgix/react-imgix/issues/469#issue-506910613) you mentioned that `Object.values` is not supported on older browsers. Care to expand on which one(s) you noticed this issue on? Or maybe provide some way to test the before/after here?

baldurh: `Object.values` isn't supported in IE11 and I've also seen this pop up in sentry for iOS 10.0.1. It was added in 10.3. 
See here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

sherwinski: After more attempts than I'd like to admit, looks like I'm unable to reproduce this issue in either IE 10 or 11.
As much as I'd like to be thorough here with regards to testing, I'm willing to move forward trusting that this babel plugin fixes this exact issue as it suggests. Any thoughts before I do so @baldurh?

baldurh: The build still contains the `Object.values` call. The plugin needs to be added to `.babelrc` for this to work. Also, I noticed that the dependency is missing from the lock file. Is updating the lock file part of the release process?

sherwinski: Ah good catch, I added the plugin to `.babelrc` and can confirm the build no longer contains `Object.values`. And yes, the lock file should have been included as well 🤦‍♂

PR #476, opened 2019-10-21
chore: refactored from findDOMNode to refs

### LINKED ISSUE #475: FindDOMNode is deprecated in StrictMode
I suppose the first question is what `onMounted` supports. It seems like it breaks React abstractions to some degree. I wasn't able to find discussion on its genesis, either.

I've been working through fixing StrictMode issues in my own project, and `findDOMNode` in this library is one of the things it's complaining about. If `onMounted` is critical functionality, it should be implemented with `ref`s. If it's not, can it be deprecated?

I'll submit a PR.

tstirrat15:
Fixes #475 

##  Description
`findDOMNode` is deprecated in `StrictMode`. It's recommended against [in the official docs](https://reactjs.org/docs/react-dom.html#finddomnode), and if the functionality that it enables is required, `ref`s are a good way of addressing the problem.

This PR refactors away from `findDOMNode` towards `ref`s.

## Changes
* Refactor `react-imgix` components to use `ref`s in favor of `findDOMNode`
* Refactor tests to use `mount` instead of `shallow`, since `shallow` doesn't create or use `ref`s
* Remove a test function that wasn't in use

## Steps to Test

See that the refactored tests are sane and passing.

## Note

I wasn't quite sure what sort of work this represented (fix/feature/chore). Please let me know if a different designation is more appropriate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "Frederick Fogerty <frederick.fogerty@gmail.com> (https://github.com/frederickfogerty)",
     "Max Kolyanov (https://github.com/maxkolyanov)",
     "Sherwin Heydarbeygi <sherwin@imgix.com> (https://github.com/sherwinski)",
-    "Baldur Helgason <baldur.helgason@gmail.com> (https://github.com/baldurh)"
+    "Baldur Helgason <baldur.helgason@gmail.com> (https://github.com/baldurh)",
+    "Tanner Stirrat <tstirrat@gmail.com> (https://github.com/tstirrat15)"
   ],
   "license": "ISC",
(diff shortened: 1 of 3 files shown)


tstirrat15:
Good point. I can refactor towards a callback ref if that'd be preferable.

frederickfogerty:
Thanks for the reply @tstirrat15. @sherwinski and @jayeb will reply here soon with a strategy 👍. Thanks for the PR btw 🙌

tstirrat15:
Done

frederickfogerty:
👏 👏 👏

PR #496, opened 2019-11-13
fix: prevent overwriting htmlAttributes.ref

### LINKED ISSUE #494: htmlAttributes ref function not called
**Before you submit:**

- [x] Please search through the existing issues (both open AND closed) to see if your issue has been discussed before. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please ensure the problem has been isolated and reduced. This link explains more: http://css-tricks.com/6263-reduced-test-cases/

**Describe the bug**
Since the 9.0 release, `ref` functions passed via `htmlAttributes` are no longer working.

**To Reproduce**

```js
const handleRef = img => {
  // this should be called after the img is rendered in the DOM
}

return (
  <Imgix
    ...
    htmlAttributes={{ ref: handleRef }}
  />
}
```

**Expected behaviour**
As per the docs, the ref function passed in `htmlAttributes` should be called after the image is loaded in the DOM with a reference to the image html element. 

**Information:**

- react-imgix version: v9.0
- I saw there was some work on ref handling for this release so I'm hoping this is straightforward to diagnose, but if not I can try to set up a repro. Thanks!

sherwinski:
Fixes #494 
Refactoring work completed in #475 inadvertently introduced a regression where we are overwriting a `ref` passed via `htmlAttributes`. This PR better handles this situation while accounting for refs that are either passed as a callback or an object.

### DIFF
--- a/src/react-imgix.jsx
+++ b/src/react-imgix.jsx
@@ -220,20 +220,34 @@ class ReactImgix extends Component {
       defaultAttributeMap,
       this.props.attributeConfig
     );
-    const childProps = Object.assign(
-      {},
-      {
-        [attributeConfig.sizes]: this.props.sizes,
-        className: this.props.className,
-        width: width <= 1 ? null : width,
-        height: height <= 1 ? null : height,
-        [attributeConfig.src]: src,
-        ref: el => {
-          this.imgRef = el;
+    const setParentRef = ref => {
+      if (!ref) {
+        return;
+      }
+
+      // assign ref based on if it's a callback vs object
+      if (typeof ref === "function") {
+        ref(this.imgRef);
+      } else {
+        ref.current = this.imgRef;
+      }
+    };
+    const childProps = Object.assign({}, this.props.htmlAttributes, {
+      [attributeConfig.sizes]: this.props.sizes,
+      className: this.props.className,
+      width: width <= 1 ? null : width,
+      height: height <= 1 ? null : height,
+      [attributeConfig.src]: src,
+      ref: el => {
+        this.imgRef = el;
+        if (
+          this.props.htmlAttributes !== undefined &&
+          "ref" in this.props.htmlAttributes
+        ) {
+          setParentRef(this.props.htmlAttributes.ref);
         }
-      },
-      this.props.htmlAttributes
-    );
+      }
+    });
     if (!disableSrcSet) {
       childProps[attributeConfig.srcSet] = srcSet;
     }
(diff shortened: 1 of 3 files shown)


sherwinski:
@jaredloson Feel free to take a look and leave feedback as well.

src/react-imgix.jsx
```diff
@@ -220,14 +220,20 @@ class ReactImgix extends Component {
...
+      {
+        [attributeConfig.sizes]: this.props.sizes,
+        className: this.props.className,
+        width: width <= 1 ? null : width,
+        height: height <= 1 ? null : height,
+        [attributeConfig.src]: src,
+        ref: el => {
+          this.imgRef = el;
+        }
+      },
```
sherwinski: The downside to this is that users could potentially overwrite any of the other properties included in this object blob via `htmlAttributes` if not careful. To prevent this, we could keep the original order but pull out the `ref` assignment before `htmlAttributes`. Thoughts?

test/unit/react-imgix.test.jsx
```diff
@@ -166,6 +166,7 @@ describe("When in image mode", () => {
       />
     );
 
+    expect(onMountedSpy.called);
```
frederickfogerty: This is a redundant assertion.

test/unit/react-imgix.test.jsx
```diff
@@ -309,6 +310,56 @@ describe("When in <source> mode", () => {
...
+          htmlAttributes={htmlAttributes}
+        />
+      );
+
+      expect(sut.props()["htmlAttributes"]["data-src"]).toEqual(
+        htmlAttributes["data-src"]
+      );
+    });
+
+    it("attaches a ref via htmlAttributes", () => {
```
frederickfogerty: We should add a test here for when `htmlAttributes.ref` is used, that onMounted is also called.

src/react-imgix.jsx
frederickfogerty: For problems like these I like to go with the second option you suggested.

src/react-imgix.jsx
```diff
@@ -220,14 +220,20 @@ class ReactImgix extends Component {
...
+      this.props.htmlAttributes
+    );
```
frederickfogerty: I'm going to go out on a limb here and say that this is not the change we want. The problem here is that our own internal ref is not set when someone overrides the ref.

Instead, I think we should combine the ref handler into one handler. Something like this:

```jsx
ref: el => {
  this.imgRef = el; // handle internal ref
  
  setParentRef(this.props.htmlAttributes.ref); // handle ref from htmlAttributes
}

// elsewhere
const setParentRef = (ref) => {
  if (!ref) { return }

  // callback refs
  if (typeof ref === 'function') {
    ref(element);
  }

  // "modern" refs
  ref.current = element;
}
```

`setParentRef` needs to be a function that will handle `undefined`, and refs in the legacy and modern styles. You can find some more information on this, and various implementations, here: https://github.com/facebook/react/issues/13029. Be wary not to use hooks as that would violate our React version support range.

This package could also be useful: https://www.npmjs.com/package/@seznam/compose-react-refs

sherwinski: Awesome thanks. If I understood your points correctly, I think it should now be able to account for each of these cases. Give it another look over and let me know.

src/react-imgix.jsx
```diff
@@ -331,12 +351,32 @@ class SourceImpl extends Component {
...
+        return;
+      }
+
+      // assign ref based on if it's a callback vs object
+      if (typeof ref === "function") {
+        ref(this.sourceRef);
+      } else {
+        ref.current = this.sourceRef;
+      }
+    };
```
frederickfogerty: I think we could keep this code more DRY and performant by making the signature `const setParentRef = (parentRef, el) =>` so that we don't have to redefine it in every render method

test/unit/react-imgix.test.jsx
```diff
@@ -309,6 +309,75 @@ describe("When in <source> mode", () => {
...
+          sizes="100vw"
+          onMounted={onMountedCallback}
+          htmlAttributes={{ ref: htmlAttrCallback }}
+        />
+      );
+
+      expect(htmlAttrCallback.called).toBeTruthy();
+      expect(htmlAttrCallback.callCount).toEqual(1);
+      expect(onMountedCallback.called).toBeTruthy();
+      expect(onMountedCallback.callCount).toEqual(1);
```
frederickfogerty: nit: `called` and `callCount` are overlapping assertions - we can just drop the `.called` assertion as the `.callCount` tests for the same thing.

test/unit/react-imgix.test.jsx
```diff
@@ -309,6 +309,75 @@ describe("When in <source> mode", () => {
...
+      sut = mount(
+        <Source
+          src={"https://mysource.imgix.net/demo.png"}
+          sizes="100vw"
+          htmlAttributes={{ ref: callback }}
+        />
+      );
+
+      expect(callback.called).toBeTruthy();
+      expect(callback.callCount).toEqual(1);
```
frederickfogerty: nit: same comment as below

test/unit/react-imgix.test.jsx
sherwinski: Ah forgot to take those out. Thanks

Issue #285, opened 2019-02-06
Srcsets, targetwidths, and server-side rendering


tstirrat15:
- [x] Please search through the existing issues (both open AND closed) to see if your question has already been discussed. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue

**Question**
I saw  in [another issue](https://github.com/imgix/react-imgix/issues/71#issuecomment-276523785) that one of the goals of react-imgix is to generate a srcset that corresponds pretty well to the rendered size of the image. This works well if the code is first being executed on the frontend and is already in a browser context, but it seems that the backend behavior is to grab [every size referenced in targetWidths.js](https://github.com/imgix/react-imgix/blob/master/src/targetWidths.js) and generate a srcset for every width. In our experience, that makes for a whole lot of markup, with sizes ranging from hundreds of pixels to thousands of pixels that aren't always relevant.

* Is there a reason for the particular granularity specified in `targetWidths`, or could those be modified?
* Are there ways to give hints to the server render about the user's viewport in a way that would produce a smaller srcset?
* For sizes, we mostly fill the width of the parent container and use `100vw` as our size prop. Is there something that we could do differently there that would produce smaller srcsets?

jayeb:
Hi Tanner, thanks for the ticket.

The issue you're referencing is two years old and that answer, while it was accurate back then (5.2.0, I believe), no longer reflects the behavior of the library.

In 8.0.0 we moved from a "measure the DOM and download the best-sized image to fit" approach to a "generate a robust `srcset` attribute and let the browser do the work" approach. This has a number of benefits:

- It works just as well server-side as it does on the browser-side.
- It's more performant for the client, since it doesn't rely on timers or events to select images.
- It's more performant in terms of download speed, since it better leverage's imgix's cache of rendered images.

Thus, the behavior you're seeing on the backend is now also the behavior on the front-end: we simply use imgix's parameters to generate a massive "menu" of sizes for the browser to choose from and list them all out in the `srcset` attribute. The first version of this behavior was somewhat naive, in retrospect--we generated our list of sizes by combining a (somewhat outdated) list of common device and screen sizes with a stepped list of dimensions from 100 to 4000.

[...]

PR #528, opened 2019-12-11
feat: added `targetMaxWidth` option

beldur:
## Description

This pull request adds the the `targetMaxWidth` option to limit the max `srcSet` size that will be automatically generated.

Fixes issue: #285


README.md
```diff
@@ -481,6 +481,10 @@ By default this component adds a parameter to the generated url to help imgix wi
 
 Any other attributes to add to the html node (example: `alt`, `data-*`, `className`).
 
+##### targetMaxWidth :: number
+
+The maximum size for the auto generated `srcSet`.
```
frederickfogerty:
```suggestion
By default this library generates srcset widths up to 8192px wide. This can be reduced by setting `targetMaxWidth` to a number lower than 8192, e.g. `targetMaxWidth={4000}`.
```

ericdeansanchez:
@beldur hey 👋 thanks so much for opening this PR! Quite a few things have changed since the opening of this PR. Namely, the propagation of custom srcset generation throughout our core-level SDKs.

We've been working on a set of features that address this issue (and others like it). The kind of functionality you seek has been implemented in our [imgix-core-js]() codebase.

We plan on working these features into `react-imgix`. The new features will give you **complete control over the srcset attribute string** that gets generated.

These features allow you to:
* define a completely custom [width-range](https://github.com/imgix/imgix-core-js#minimum-and-maximum-width-ranges) by setting min _and_ max widths **and**
* tune width-ranges with a custom [`widthTolerance`](https://github.com/imgix/imgix-core-js#width-tolerance) to control the granularity of the srcset image widths within the range (i.e. between min and max widths) **and**
* supply your own widths [directly](https://github.com/imgix/imgix-core-js#custom-widths), eg.

```js
{ widths: [320, 440, 640, 720] }
```

@tstirrat15 Hello, I noticed your reply on an old issue that this PR references. I just wanted to "at" you here so you could hopefully benefit from the above context as well.

beldur:
Thanks a lot!

tremby:
I came here looking for this. Any news on it? It doesn't look like this feature has been added unless I'm looking in the wrong places.

At the moment react-imgix is putting entries in the srcset way up to 8192 pixels wide. If my image is only 1000 pixels wide, a browser on a display with a lot of pixels is going to be requesting images way bigger than the source, and Imgix is going to pointlessly upscale the image, wasting the user's time and data.

I can mitigate this by adding `fit=max` so at least it's never upscaling, but the browser will still download what it thinks are going to be larger versions of the image if the viewport size is increased, say, from 1400px wide to 2000px wide. But for all I know browser implementations may then have unexpected behaviour when the received image isn't actually the width the srcset value said it was going to be.

I should definitely be able to tell Imgix what I know to the the largest available width of the image so it doesn't falsely promise the user that larger versions are available.

To me the lack of this option seems like a very important bug. It's potentially a lot of wasted CPU cycles on both the Imgix side and the client side, wasted cache space on both sides, wasted network traffic, and wasted transfer time.

The lack of ability to make the gap larger between available sizes is another issue -- the smaller the gap is, the less likely it is that the size the browser decides to download has already been generated and cached, meaning a more sluggish experience on average.

sherwinski:
Continuing this conversation over at to #723

PR #592, opened 2020-03-31
fix(background): fortify `hasDOMDimensions` check for null height

### LINKED ISSUE #591: <Background /> with a width (and `top`) but no height causes Imgix 400
Currently the `Background` component is set up in such a way where if the image "isn't ready", then it doesn't attempt a render.

This is great, except for the "is ready" check, which checks `top != null` of the bounding client rect. However, a very common flow will be that the DOM _node_ is ready (i.e. it is in the DOM, and thus has a truthy `top`), but the CSS/layout hasn't finished rendering fully, such that the DOM still has no width/height.

This results in react-imgix trying to use a `height` of `0`, which then ends up giving `const ar = contentRect.bounds.width / contentRect.bounds.height;` an `ar` of `NaN`, and ends up with imgix URLs of e.g. `/blah/foo.jpeg?ar=3%3A4&fit=crop&h=NaN&q=80&w=100`

This then has react-imgix hit a 400

stephencookdev:
## Description

This PR changes the `hasDOMDimensions` to check for DOM `width` and `height`, rather than just the `top` — which will exist in cases where we still don't want to consider the DOM dimensions "usable".

This should fix #591 — which should, I imagine, stop Imgix from getting quite a few 400s

## Checklist

Please use the checklist that is most closely related to your PR _(you only need to use one checklist, and you can skip items that aren't applicable or don't make sense)_:

### Bug Fix

- [x] Add some [steps](#steps-to-test) so we can test your bug fix

N/A:
- [x] (done since N/A) Update the readme
- [x] (done since N/A) Update or add any necessary API documentation

## Steps to Test

The new unit test should be sufficient.

But if you want to test manually, too, then create a `<div style={{ width: 100, height: 0 }} /><Background /></div>` element

Related issue: #591

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "Max Kolyanov (https://github.com/maxkolyanov)",
     "Sherwin Heydarbeygi <sherwin@imgix.com> (https://github.com/sherwinski)",
     "Baldur Helgason <baldur.helgason@gmail.com> (https://github.com/baldurh)",
-    "Tanner Stirrat <tstirrat@gmail.com> (https://github.com/tstirrat15)"
+    "Tanner Stirrat <tstirrat@gmail.com> (https://github.com/tstirrat15)",
+    "Stephen Cook <stephen@stephencookdev.com> (https://github.com/stephencookdev)"
   ],
   "license": "ISC",
   "bugs": {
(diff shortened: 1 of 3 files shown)


sherwinski:
@all-contributors add @stephencookdev for code and bugs

PR #605, opened 2020-04-13
docs: make formatting fixes for Markdown quirks

jayeb:
## Description

This PR makes a number of small changes to `README.md`, to account for some Markdown quirks.

Because this README file is intended to render in two places (Here on GitHub, and on [our Docs site](https://docs.imgix.com/libraries/react-imgix)), we need to be extra careful with our Markdown syntax. GitHub's Markdown rendering is extremely forgiving, but the library we use to power Docs ([showdown.js](https://github.com/showdownjs/showdown)) adheres much more closely to the official Markdown spec, which creates some inconsistencies that, for now, need to be fixed at the content layer.

This PR includes three such changes:

- The indentation on the table of contents has been adjusted to match the official Markdown spec of "4 spaces for nested list indentation". This unfortunately means we can no longer simply drop in the output from `markdown-toc` for this section, but adjusting indentation "manually" is fortunately not a terrible chore.
- The nested codeblock under the "7.x to 8.0" upgrade guide has been fixed so that it and its sibling paragraph are both indented by 4 spaces, making them proper children of the parent list item, in the eyes of the official Markdown spec.
- The BrowserStack logo image located in the "Browser Support" section has been updated to use an absolute path, which ensures the image will render properly outside of GitHub contexts.

### Documentation

- [x] The PR title is in the [conventional commit](#conventional-commit-spec) format: e.g. `chore(readme): updated documentation for ___`


README.md
```diff
@@ -16,7 +16,7 @@
 
 <!-- /ix-docs-ignore -->
 
-<!-- NB: Run `npx markdown-toc README.md -i --maxdepth 4` to generate TOC :) -->
+<!-- NB: Run `npx markdown-toc README.md -i --maxdepth 4` to generate TOC, then manually adjust indentation to 4 spaces :) -->
```
frederickfogerty, line 19: No change needed, but this command will also do the same thing without us having to manually re-indent every time: `npx markdown-toc README.md --maxdepth 4 | sed -e 's/[[:space:]]\{2\}/    /g'`. All we have to do is manually insert the TOC, it won't insert it automatically anymore.


```suggestion
<!-- NB: Run `npx markdown-toc README.md --maxdepth 4 | sed -e 's/[[:space:]]\{2\}/    /g'` to generate TOC, and copy the result from the Terminal to replace the TOC below :) -->
```

PR #673, opened 2020-08-07
chore: change refs of master to main

frederickfogerty:
This PR changes any references from the master branch to the main branch, now that it is the default.

### DIFF
--- a/src/constructUrl.js
+++ b/src/constructUrl.js
@@ -1,6 +1,6 @@
 /*
 Copyright © 2015 by Coursera
-Licensed under the Apache License 2.0, seen https://github.com/coursera/react-imgix/blob/master/LICENSE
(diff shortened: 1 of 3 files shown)


src/constructUrl.js
```diff
@@ -1,6 +1,6 @@
 /*
 Copyright © 2015 by Coursera
-Licensed under the Apache License 2.0, seen https://github.com/coursera/react-imgix/blob/master/LICENSE
+Licensed under the Apache License 2.0, seen https://github.com/coursera/react-imgix/blob/main/LICENSE
```
sherwinski, line 3: Looks like this one should be reverted

frederickfogerty: Oh you're right - I actually picked this up locally but didn't push my commit. Fixing now.

PR #679, opened 2020-08-13
chore: update some more refs to master

frederickfogerty:
This PR follows up on the branch changes with some changes that were missed initially.


sherwinski:
So many references 😵

PR #686, opened 2020-08-27
fix: use js-base64@3.4.5 sans eval

### LINKED ISSUE #565: js-base64 uses eval, leading to CSP violations
When using react-imgix on a site with Content Security Policy, one of react-imgix's dependencies triggers CSP violations by using eval. 

From Chrome dev tools:
```
base64.js:30 [Report Only] Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "default-src https: 'unsafe-inline'". Note that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.
```

The offending line looks like it's feature detecting nodejs:
https://github.com/dankogai/js-base64/blob/e8a9a09edaf79fddee3623d97421151dcbd384c9/base64.js#L30

Further, a base64 library should not be needed in a browser environment, which has `atob()` and `btoa()` available. The fallback library could be loaded in a node environment if or native base64 support were not available.

A build of react-imgix that does not use eval.

**Describe alternatives you've considered**
- Forking react-imgix to meet our needs. This is undesirable because we'd lose out on updates and other enhancements in react-imigx, and have to maintain additional 3rd party libraries.

- Bundling react-imgix but stubbing out js-base64. This is still undesirable because we'd have to modify other parts of our build and dependency management chain to accommodate this exception.

ericdeansanchez:
The purpose of this PR is to address #565 wherein the older version of js-base64
used `eval` leading to a CSP violation. This PR updates the dependency to use
version `3.4.5`. Since `3.3` [js-base64](https://github.com/dankogai/js-base64#heads-up-switch-to-typescript-since-version-33) has switch to TS and no longer contains
references to `eval()`.

In #565 a desired solution was to offer a build of `react-imgix` that does not use
`eval()`, and this does that.

Perhaps there's an opportunity to use cypress here to test for CSP violations?

Closes #565

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3760,11 +3760,11 @@
       }
     },
     "@babel/runtime": {
-      "version": "7.6.3",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.6.3.tgz",
-      "integrity": "sha512-kq6anf9JGjW8Nt5rYfEuGRaEAaH1mkv3Bbu6rYvLOpPh/RusSJXuKPEAoZ7L7gybZkchE8+NV5g9vKF4AGAtsA==",
+      "version": "7.11.2",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.11.2.tgz",
+      "integrity": "sha512-TeWkU52so0mPtDcaCTxNBI/IHiz0pZgr8VEFqXFtZWpYD08ZB6FaSwVAS8MKRQAP3bYKiVjwysOJgMFY28o6Tw==",
       "requires": {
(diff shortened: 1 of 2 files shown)


frederickfogerty (changes requested):
Hey @ericdeansanchez, thanks for opening this PR. Unfortunately it looks like this version of js-base64 doesn't work on IE 11! 😢

ericdeansanchez:
Oof, yeah I see that. It was worth a shot :sweat_smile: Hmmm, okay I have some thoughts on how to proceed here. I'm closing this PR and will open another (possibly one that doesn't depend on `js-base64` at all.

PR #735, opened 2020-12-04
ci: update supported node versions

sherwinski:
Adds [currently supported](https://nodejs.org/en/about/releases/) LTS node versions to CI.

TODO: remove EOL'd versions in a future major release

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,8 @@
   },
   "peerDependencies": {
     "prop-types": ">=15.5.6",
-    "react": ">=15",
-    "react-dom": ">=15"
+    "react": "15.x || 16.x || 17.x",
+    "react-dom": "15.x || 16.x || 17.x"
   },
   "dependencies": {
     "js-base64": "~2.6.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ branches:
     - /release.*/
 
 node_js:
-  - "9"
+  - lts/*
+  - stable
 
 script:
   - travis_retry npm test


frederickfogerty (changes requested):
I don't think we should add `10` here since it's already EOL'd. Maybe it'd be better to use the lts + stable strategy from the other PR. Also I'm not sure how much we get out of testing multiple node versions, since we have browserstack tests, and the code is transpiled before release anyway.

sherwinski:
Based on the [react website](https://reactjs.org/versions/), react versions 15 through 17 are still supported. I've updated this project's peer dependencies to call this out more explicitly. Checking the tip of each of those branches [[1](https://github.com/facebook/react/blob/v15.7.0/package.json#L100)][[2](https://github.com/facebook/react/blob/v16.14.0/package.json#L100)][[3](https://github.com/facebook/react/blob/v17.0.1/package.json#L100)] shows that node versions 8 through 14 are supported, but I think it's enough to test against lts and stable versions here.

package.json
```diff
@@ -51,8 +51,8 @@
   },
   "peerDependencies": {
     "prop-types": ">=15.5.6",
-    "react": ">=15",
-    "react-dom": ">=15"
+    "react": "15.x || 16.x || 17.x",
+    "react-dom": "15.x || 16.x || 17.x"
```
frederickfogerty, line 55: Although I understand the reasoning behind explicitly calling out the React versions, I think this actually adds more of a headache, since if React releases a new version, we won't be compatible with it and will need to be quick on the ball here. I think we can still keep this change but add something like `>17` so that new versions are automatically supported. e.g. 

```suggestion
    "react": "15 || 16 || 17 || >17",
    "react-dom": "15 || 16 || 17 || >17"
```

frederickfogerty (changes requested):
Sounds like good enough reasoning to me.

frederickfogerty (approved):
Looks good!

PR #746, opened 2021-01-22
config: semantic release

frederickfogerty:
This PR adds semantic release to this library. This config is pretty standard, and I tested a deploy out locally as a dry run and everything went fine.


sherwinski (approved):
LGTM 👍

PR #764, opened 2021-03-16
config: use @imgix/browserslist-config for browser testing

sherwinski:


frederickfogerty (approved):
👏 very nice! Glad to have this in

PR #776, opened 2021-04-01
feat: integrate js-core into react-imgix

### LINKED ISSUE #763: <Background> generating parameters twice for width and height
Based on the code I had to send imgixParams{ w: X, h: Y } to use the forced values.

But this would result in two parameters w and h appended to the URL.

Expected result: Only one parameter should be sent for a given key.

ericdeansanchez:
The purpose of this PR is to integrate `@imgix/js-core` into
`react-imgix`. First, `targetWidths` has been reimplemented in terms of
the js-core client. Second, `buildSrc` has been implemented in terms of
`ImgixClient.buildSrcSet`. Lastly, a temp fix was made for the
`Background` component. Prior to this PR, the flow of the code/params
caused duplicate param key-value pairs to be written into query strings.

Ideally, we want to avoid getting into this kind of situation
altogether. However, doing that will require a bit more reworking,
refactoring, and (possibly || probably) breaking changes.

Closes #763


src/react-imgix.jsx
```diff
@@ -109,68 +110,97 @@ function buildSrc({
...
-    src,
-    srcSet,
-  };
+    if (width) {
+      params["w"] = width ? width : w;
+    }
+
+    if (height) {
+      params["h"] = height || h;
+    }
```
luqven, line 176: Nit. Why are we assigning conditionally values differently here? Ie `bar = foo ? foo : f`  vs `bar = foo || f`. Both `width` and `height` are declared in this closure no? So we can just use `bar = foo || f` for both w and h.

src/targetWidths.js
```diff
@@ -1,18 +1,3 @@
...
-    resolutions.push(ensureEven(prev));
-    prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
-  }
-
-  resolutions.push(MAX_SIZE);
-  return resolutions;
-}
-
-export default targetWidths();
+export default ImgixClient.targetWidths();
```
luqven, line 3: nice!  🎉

luqven (approved):
LGTM 👍🏼

src/targetWidths.js
ericdeansanchez: Yeah, I think we might change this setup (probably don't need a whole file to do this now).

src/react-imgix.jsx
ericdeansanchez: Ooh, yeah. I had experimented with something else here and left it inconsistent. I can change this.

src/react-imgix-bg.jsx
```diff
@@ -87,14 +86,13 @@ const BackgroundImpl = (props) => {
 
   const renderedSrc = (() => {
     const [rawSrc, params] = extractQueryParams(src);
+    const allParams = { ...params, ...imgixParams, dpr };
```
sherwinski, line 89: 👍

src/react-imgix.jsx
```diff
@@ -109,68 +110,97 @@ function buildSrc({
...
-        srcFn = buildDprSrcWithQuality;
-      } else if (disableQualityByDPR) {
-        srcFn = buildDprSrcWithoutQuality;
+    for (const k in urlParams) {
+      if (PARAM_EXPANSION[k]) {
+        params[PARAM_EXPANSION[k]] = urlParams[k];
+      } else {
+        params[k] = urlParams[k];
       }
+    }
```
sherwinski, line 128: Might be out of scope here but just a general thought: Would it be worth packaging this functionality up in some sort of helper, since it appears here as well: https://github.com/imgix/react-imgix/pull/776/files#diff-11b2734ddfc432c9651ade2ada0deb5f2201e6e88e3fcc271ebbcac05de8c82fR118-R124

src/react-imgix.jsx
```diff
@@ -109,68 +110,97 @@ function buildSrc({
...
+
+    const [scheme, rest] = rawSrc.split("://");
+    const [domain, ...pathComponents] = rest.split("/");
+    let useHTTPS = scheme == "https";
+
+    const client = new ImgixClient({
+      domain: domain,
+      useHTTPS: useHTTPS,
+      includeLibraryParam: false,
+    });
```
sherwinski, line 186: I noticed `ImgixClient` gets setup several times throughout this component. I'm honestly not sure if it has any bearing on performance, but I get the feeling it should be pulled up more towards the top level of the component to avoid repeated work.

src/targetWidths.js
sherwinski: chefskiss.jpg

src/react-imgix.jsx
ericdeansanchez: Definitely, I think I should do that next. Right now this pattern exists in three different places, I'll wrap it up into a helper.

src/react-imgix.jsx
ericdeansanchez: I don't think it impacts performance because we only setup a client once either way (once in each block of the if-else blocks). The if-else blocks are kind of long, so I made each do one job (keeping things scoped/encapsulated into each block) then each block returns it's result. In general, I try no to use a single, top-level definition, through multiple long blocks of code (this is defensive, i.e. to prevent reassignment or some mutation that might make debugging difficult later on).

That said, I think we can shorten things up a lot here. Now that this initial bit has been looked over I'll make another pass over things refactoring/cleaning up.

src/react-imgix-bg.jsx
ericdeansanchez: I can definitely employ more object spread to shorten some other bits of code up here too.

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+      height
+    );
+
+    const src = constructUrl(rawSrc, srcOptions);
+
+    return { src, srcSet };
+  } else {
+    warnInvalidAspectRatio(imgixParams.ar || "", config);
+
+    const { w, h, ...urlParams } = srcOptions;
```
frederickfogerty, line 213: It seems like this destructuring is unnecessary since this behaviour is handled by `fixedSize`?

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+    const src = constructUrl(rawSrc, srcOptions);
+    const srcSet = buildSrcSet(rawSrc, { ...urlParams });
```
frederickfogerty, line 215: We shouldn't need to spread here anymore 🤞

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+}
+
+function buildChildProps(obj, src, attributeConfig, width, height, refType) {
+  const childProps = {
+    ...obj.props.htmlAttributes,
+    [attributeConfig.sizes]: obj.props.sizes,
+    className: obj.props.className,
+    width: width <= 1 ? null : width,
+    height: height <= 1 ? null : height,
+    [attributeConfig.src]: src,
```
frederickfogerty, line 260: This shouldn't set `src` for `<source>` elements

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+ */
+function aspectRatioIsValid(aspectRatio) {
+  if (typeof aspectRatio !== "string") {
+    return false;
+  }
+
+  return /^\d+(\.\d+)?:\d+(\.\d+)?$/.test(aspectRatio);
+}
+
+function buildChildProps(obj, src, attributeConfig, width, height, refType) {
```
frederickfogerty, line 253: I think this function relates more to the components where it is called, so at least it should be located in that file.

It seems like a bit of a smell to be passing `this` around - seems like it could go wrong in the future if this function was to incorrectly assign to `obj`. I wonder if it's better if it's possible to make this pure, or just to deal with this logic being duplicated in the separate components. It's not super complicated so I'm not sure how much benefit there is to extract it into a common function.

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+  }
+}
+
+function buildSrcSet(rawSrc, params = {}, options = {}, width, height) {
+  const { client, pathComponents } = extractClientAndPathComponents(rawSrc);
+  const compactedParams = compactParamKeys(params, width, height);
+  return client.buildSrcSet(pathComponents.join("/"), compactedParams, options);
+}
+
+function warnInvalidAspectRatio(aspectRatio, config) {
```
frederickfogerty, line 227: Nice 👍

src/react-imgix-bg.jsx
```diff
@@ -87,14 +86,13 @@ const BackgroundImpl = (props) => {
...
     const srcOptions = {
-      ...params,
       fit: "crop",
-      ...imgixParams,
       ...(disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` }),
+      ...allParams,
```
frederickfogerty, line 93: Is this the fix related to this comment?

> Lastly, a temp fix was made for the Background component. Prior to this PR, the flow of the code/params caused duplicate param key-value pairs to be written into query strings.

src/targetWidths.js
frederickfogerty, line 3: Haha, nice! 👏

test/unit/react-imgix.test.jsx
```diff
@@ -62,10 +62,10 @@ describe("When in default mode", () => {
       const sut = shallow(<Imgix src={src} sizes="100vw" />);
       const srcset = sut.props().srcSet;
       expect(srcset).not.toBeUndefined();
-      expect(srcset.split(", ")[0].split(" ")).toHaveLength(2);
-      const aSrcFromSrcSet = srcset.split(", ")[0].split(" ")[0];
+      expect(srcset.split(",\n")[0].split(" ")).toHaveLength(2);
+      const aSrcFromSrcSet = srcset.split(",\n")[0].split(" ")[0];
       expect(aSrcFromSrcSet).toContain(src);
-      const aWidthFromSrcSet = srcset.split(", ")[0].split(" ")[1];
+      const aWidthFromSrcSet = srcset.split(",\n")[0].split(" ")[1];
```
frederickfogerty, line 68: There's probably little value testing this stuff in react-imgix anymore since it's now implemented and tested by `@imgix/js-core`

frederickfogerty (changes requested):
Overall I think this is really great and reduces the complexity of this library. Nice work! 🎉 

- `DPR_QUALITY` and `DPR_QUALITY_VALUES` should be able to be removed from `constants.js` because I think they're no longer used
- I think the changes to use `{...object}` rather than `Object.assign` are nice stylistically, but they were introduced for performance reasons, and we shouldn't change these back unless we're sure that there is no change to performance. There was a 30% increase in performance and it would be a shame to lose these gains. Context: #423 
- The title should be updated to not include a "."

src/HOCs/shouldComponentUpdateHOC.jsx
```diff
@@ -12,28 +12,26 @@ const ShouldComponentUpdateHOC = (WrappedComponent) => {
       );
 
       const customizer = (oldProp, newProp, key) => {
```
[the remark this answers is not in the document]
ericdeansanchez: Yeah, I've had similar thoughts. I kept it in because it was there before, but it's been suspect. One thing I thought about doing (maybe/not in this PR) was to implement `shouldComponentUpdate` for each component. Initially this thought came from wanting to remove the indirection added by using an HOC for this. Thanks for brining this up, this is a really good point, I can double check that this is essentially no-op/default-op behavior and we can take it out.

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
-      params,
-      imgixParams,
-      disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` }
-    )
-  );
+  return constructUrl(rawSrc, {
+    ...params,
+    ...imgixParams,
+    ...(disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` }),
+  });
```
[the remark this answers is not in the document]
ericdeansanchez: Agreed! I can't walk this back... or really I wanted to benchmark things pre/post imgix/js-core integration; I can continue ahead with that work.. it might be more necessary than not at this point. I have seen both `Object.assign` and `Object spread` coming through perf tools, let me try to get some numbers.

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
+    ...params,
+    ...imgixParams,
+    ...(disableLibraryParam ? {} : { ixlib: `react-${PACKAGE_VERSION}` }),
+    ...(fixedSize && height ? { height } : {}),
+    ...(fixedSize && width ? { width } : {}),
+  };
+
+  if (disableSrcSet) {
+    const src = constructUrl(rawSrc, srcOptions);
+    return { src, src };
```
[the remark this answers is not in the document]
ericdeansanchez: Yeah, it looks weird. Maybe `{ src, srcSet: src }` would be better (more explicit). Though, what I'd really like to do is to break `buildSrc` up so that it only returns a srcset. The reason being that I'd like to change up the flow within components (i.e. conditionally generating a srcset only when it's needed and not disabled). I'd also like to rearrange/"hoist" some logic up so that we're not repeating work in later steps. I'd like to change this, bug/mistake or not. I'll look into this.

src/constructUrl.js
ericdeansanchez: Yeah, that's what I thought too. However, if you remove `w` and `h` you reveal a subtle bug in the implementation logic. Before these changes I think the original impl was destructuring to remove `w` and `h` from the options and only propagate `urlParams` down through the rest of the logic.

I think there may need to be some three-way-merging needed here? Something like "combine/resolve which height and width to use given those from props, ixParams, and from extracting from a src."

src/constructUrl.js
ericdeansanchez: Ty ty 😄

test/unit/react-imgix.test.jsx
ericdeansanchez: I agree. I wanted to at least demo "hey, this works kind of like it did before." I wonder what work I could do to distill down the test suite now.

src/HOCs/shouldComponentUpdateHOC.jsx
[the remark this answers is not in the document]
frederickfogerty: SGTM

src/constructUrl.js
```diff
@@ -113,49 +114,176 @@ var DEFAULT_OPTIONS = Object.freeze({
...
-    url += key + "=" + encodeURIComponent(val) + "&";
+  if (height) {
+    params["h"] = height;
   }
+  return params;
+}
+
+function extractClientAndPathComponents(src) {
+  const [scheme, rest] = src.split("://");
+  const [domain, ...pathComponents] = rest.split("/");
```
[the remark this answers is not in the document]
ericdeansanchez: Won't this add encoding to the path though? Or maybe I don't understand how I'd use the built-in interface?

src/constructUrl.js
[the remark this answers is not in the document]
ericdeansanchez: ✅

src/constructUrl.js
ericdeansanchez: ✅

src/constructUrl.js
ericdeansanchez: Oof, yes, good catch. Thanks!

src/constructUrl.js
ericdeansanchez: Yeah, I think it smells a bit too, but I wanted to try to extract/distill some of the structure of what's going on here. I'll look into refactoring this.

ericdeansanchez:
Closing in favor of #780

src/constructUrl.js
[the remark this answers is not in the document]
frederickfogerty: Yeah I think you're right

PR #780, opened 2021-04-14
feat: integrate @imgix/js-core into react-imgix

### LINKED ISSUE #763: <Background> generating parameters twice for width and height

ericdeansanchez:
The purpose of this PR is to integrate `@imgix/js-core` into
`react-imgix`. First, `targetWidths` has been reimplemented in terms of
the js-core client. Second, `buildSrc` has been implemented in terms of
`ImgixClient.buildSrcSet`. Lastly, a bug fix was made for the
`Background` component.

Closes #763

### DIFF
--- a/src/constructUrl.js
+++ b/src/constructUrl.js
@@ -121,7 +121,7 @@ function constructUrl(src, longOptions) {
   return client.buildURL(pathComponents.join("/"), params);
 }
 
-function compactParamKeys(longOptions, width, height) {
+function compactParamKeys(longOptions) {
   const keys = Object.keys(longOptions);
   const keysLength = keys.length;
   const params = {};
@@ -134,13 +134,6 @@ function compactParamKeys(longOptions, width, height) {
     }
   }
 
-  if (width) {
-    params["w"] = width;
-  }
-
-  if (height) {
-    params["h"] = height;
-  }
   return params;
 }
 
--- a/src/constructUrl.js
+++ b/src/constructUrl.js
@@ -5,7 +5,6 @@ Licensed under the Apache License 2.0, seen https://github.com/coursera/react-im
 Minor syntax modifications have been made
 */
 
-var Base64 = require("js-base64").Base64;
 const PACKAGE_VERSION = require("../package.json").version;
 import extractQueryParams from "./extractQueryParams";
 import ImgixClient from "@imgix/js-core";
--- a/src/constructUrl.js
+++ b/src/constructUrl.js
@@ -8,6 +8,7 @@ Minor syntax modifications have been made
 var Base64 = require("js-base64").Base64;
 const PACKAGE_VERSION = require("../package.json").version;
 import extractQueryParams from "./extractQueryParams";
+import ImgixClient from "@imgix/js-core";
 
 // @see https://www.imgix.com/docs/reference
 var PARAM_EXPANSION = Object.freeze({
@@ -116,28 +117,46 @@ function constructUrl(src, longOptions) {
   if (!src) {
     return "";
   }
+  const params = compactParamKeys(longOptions);
+  const { client, pathComponents } = extractClientAndPathComponents(src);
+  return client.buildURL(pathComponents.join("/"), params);
+}
 
+function compactParamKeys(longOptions, width, height) {
   const keys = Object.keys(longOptions);
   const keysLength = keys.length;
-  let url = src + "?";
+  const params = {};
   for (let i = 0; i < keysLength; i++) {
     let key = keys[i];
-    let val = longOptions[key];
-
     if (PARAM_EXPANSION[key]) {
-      key = PARAM_EXPANSION[key];
+      params[PARAM_EXPANSION[key]] = longOptions[key];
     } else {
-      key = encodeURIComponent(key);
+      params[key] = longOptions[key];
     }
-
-    if (key.substr(-2) === "64") {
-      val = Base64.encodeURI(val);
-    }
-
-    url += key + "=" + encodeURIComponent(val) + "&";
   }
 
-  return url.slice(0, -1);
+  if (width) {
+    params["w"] = width;
+  }
+
+  if (height) {
+    params["h"] = height;
+  }
+  return params;
+}
+
+function extractClientAndPathComponents(src) {
+  const [scheme, rest] = src.split("://");
+  const [domain, ...pathComponents] = rest.split("/");
+  let useHTTPS = scheme == "https";
+
+  const client = new ImgixClient({
+    domain: domain,
+    useHTTPS: useHTTPS,
+    includeLibraryParam: false,
+  });
+
+  return { client, pathComponents };
 }
 
 function buildURLPublic(src, imgixParams = {}, options = {}) {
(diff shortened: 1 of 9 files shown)


ericdeansanchez:
For the whole checking for `!sizes`,  thing I coded to this test:

https://github.com/imgix/react-imgix/blob/9b080a62704cc62063c16b903413496bf0b2d0f1/test/unit/react-imgix.test.jsx#L662

src/constructUrl.js
```diff
@@ -5,9 +5,9 @@ Licensed under the Apache License 2.0, seen https://github.com/coursera/react-im
 Minor syntax modifications have been made
 */
 
-var Base64 = require("js-base64").Base64;
 const PACKAGE_VERSION = require("../package.json").version;
```
frederickfogerty, line 8: Let's change this to using `common/constants.js`, must have slipped through the cracks.

frederickfogerty (commented):
- Is

src/constructUrl.js
frederickfogerty, line 8: Let's change this to using `common/constants.js`, must have slipped through the cracks.

src/constructUrl.js
```diff
@@ -116,28 +116,39 @@ function constructUrl(src, longOptions) {
...
-  }
+function extractClientAndPathComponents(src) {
+  const [scheme, rest] = src.split("://");
+  const [domain, ...pathComponents] = rest.split("/");
+  let useHTTPS = scheme == "https";
+
+  const client = new ImgixClient({
+    domain: domain,
+    useHTTPS: useHTTPS,
+    includeLibraryParam: false,
```
frederickfogerty, line 148: Is `ixlib` reset somewhere else?

src/react-imgix.jsx
```diff
@@ -95,6 +88,7 @@ function buildSrc({
   disableSrcSet,
   imgixParams,
   disableQualityByDPR,
+  sizes,
```
frederickfogerty, line 91: What does sizes do here?

src/react-imgix.jsx
```diff
@@ -117,43 +111,44 @@ function buildSrc({
     srcSet = src;
   } else {
     if (fixedSize) {
-      const { q, ...urlParams } = srcOptions;
-      const constructedUrl = constructUrl(rawSrc, urlParams);
-
-      let srcFn = buildDprSrcWithQualityByDpr;
+      const { width, w, height, h, q, ...urlParams } = srcOptions;
```
frederickfogerty, line 114: No change needed, but I think it would be an improvement if these width and w values were "collapsed" down to a single value before getting to this point. IMO having both here seems like a bit of a smell

frederickfogerty (approved):
Looking pretty great! This is clean ✨ 👏

src/constructUrl.js
ericdeansanchez: Yes, the client's `ixlib` is disabled here in order to defer to react-imgix's way of setting this value.
https://github.com/imgix/react-imgix/blob/ae9bcfb773dcb92f53319cea01a1c02ba8cb51bc/src/react-imgix.jsx#L101

src/react-imgix.jsx
ericdeansanchez: I agree, what would you suggest?

src/react-imgix.jsx
ericdeansanchez: There's this test:
https://github.com/imgix/react-imgix/blob/ae9bcfb773dcb92f53319cea01a1c02ba8cb51bc/test/unit/react-imgix.test.jsx#L669

```jsx
it("responsive width should overwrite the width query parameter correctly", () => {
  sut = shallow(
    <Imgix src={"https://mysource.imgix.net/demo.png?w=333"} sizes="100wv" />
  );

  expect(sut.props().src).toContain("w=333");
  expect(sut.props().srcSet).not.toContain("w=333"); // <---- this check will fail if we don't check if sizes was passed.
});
```

I (very) briefly touched on this in the commit message:
>Prior to this PR, there seems to have been a little error when handling height and width values in the presence of sizes.

It seems like we want different behavior when `sizes` is passed, so that get's explicitly handled now. However, tbh this feels like a bit of a hack, so I'm open to any suggestions here.

frederickfogerty: Okay, interesting. I'm unaware of any context on this "feature" so I'm struggling a bit to understand why we would want to change this behaviour based on whether sizes was passed or not. Do we do something along the lines of checking if sizes is passed to determine whether this is a responsive image or not.

Use your best judgement here because you have the most context, but this doesn't seem like something we need to explicitly "support"

src/react-imgix.jsx
frederickfogerty: I would suggest having a "props processing" stage that occurs at the start of the component render function. This stage would take the component props (or at first it could just take the imgix params), and transform this down to a consistent set of params with all aliases flattened to the un-aliased values. For example:

```jsx
const ReactImgix = props => {
	// Process imgix params
	const imgixParams = sanitizeImgixParams(props.imgixParams);

	// Now in the rest of the function, these imgix params can be passed around safely
	const src = buildUrl(path, imgixParams);

	return <img src={src} />;
}


const ALIAS_MAP = {width: 'w', height: 'h', /* ... */ };
const sanitizeImgixParams = imgixParams => Object.entries(imgixParams).reduce((acc, [key, value]) => {
	const targetKey = key in ALIAS_MAP ? ALIAS_MAP[key] : key;
	acc[targetKey] = value;
	return acc
}, {});
```

This naive implementation doesn't make any guarantees about the order that the values will be un-aliased/applied, but I think that's ok.

A more in-depth implementation here could handle the interactions between the imgix params that are currently able to be set with props and also inside `imgixParams`.

src/react-imgix.jsx
ericdeansanchez: Yeah, I thought this bit was interesting as well. I think the prior implementation shows that this is somewhat related to the destructuring below. This line is from main:

https://github.com/imgix/react-imgix/blob/0d52511b4dddfc0d314988d12600a38751cb857f/src/react-imgix.jsx#L138

I think what was going on here is that `w width height` are all removed/excluded from the `imgixParams`. So when this test runs:

https://github.com/imgix/react-imgix/blob/ae9bcfb773dcb92f53319cea01a1c02ba8cb51bc/test/unit/react-imgix.test.jsx#L669

Specially this portion:

https://github.com/imgix/react-imgix/blob/ae9bcfb773dcb92f53319cea01a1c02ba8cb51bc/test/unit/react-imgix.test.jsx#L675

The test passes under the old implementation, but fails under the new because of the way width/height params are being passed and not being excluded. This test seems to be, "when an initial width is given, but `sizes` is also given, omit the initial width and create a width-based srcset."

```jsx
<Imgix src={"https://mysource.imgix.net/demo.png?w=333"} sizes="100wv" />
```

The only way to tell from the above that the srcset URLs should not contain a `w=333`(but should contain 31 other widths) is the `sizes` prop.

src/react-imgix.jsx
ericdeansanchez: Yeah, I think we really need to do some work to improve the hygiene around props, width, height, etc. There's the issue of duplicate params in Background URLs, plus there's the issues above wrt sizes, so this could all use some work, I agree.

PR #781, opened 2021-04-14
config(semantic): fix package version replacement

frederickfogerty:
Looks like we are using double quotes rather than single quotes in this lib, which caused the semantic release step to fail.

I can't easily guarantee this works, will have to merge to main to see if it does.

### DIFF
--- a/package.json
+++ b/package.json
@@ -232,8 +232,8 @@
               "files": [
                 "src/constants.js"
               ],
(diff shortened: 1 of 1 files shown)


sherwinski (approved):
Let's give it a shot 👍

frederickfogerty:
Looks like it worked ✌️

PR #782, opened 2021-04-19
fix: re-render Background only if a larger image is needed

### LINKED ISSUE #387: Using `<Background />` tanks my CPU

ericdeansanchez:
The purpose of this PR is to keep as much of the original behavior of the
`Background` component as possible while curbing the number of times
this component re-renders. Prior, this component re-renders an "infinite"
number of times. Really, it doesn't stop re-rendering.

Now, this component re-renders on a more appropriate interval. Prior,
this component would re-render beginning with the first render. This
means that if you were to render a `Background` component without
resizing the browser at all it would render a few hundred times in the span
of a couple seconds. Now, there are three initial renders.

This rendering behavior is expected [given the behavior of react-measure
in this case](https://github.com/souporserious/react-measure#onresize-proptypesfunc).

Prior (and ignoring the continuous rendering behavior), this component
could re-render if height/width changed. Now, this component will
only be re-rendered if height/width have **_increased_**.

Closes #387

### DIFF
--- a/src/react-imgix-bg.jsx
+++ b/src/react-imgix-bg.jsx
@@ -5,6 +5,7 @@ import constructUrl from "./constructUrl";
 import extractQueryParams from "./extractQueryParams";
 import findClosest from "./findClosest";
 import targetWidths from "./targetWidths";
+import { shallowEqual } from "./common";
 
 const noop = () => {};
 
@@ -46,7 +47,17 @@ class BackgroundImpl extends React.Component {
       return true;
     }
 
-    return false;
+    // If we made it here, we need to check if the "top-level"
+    // props have changed (e.g. disableLibraryParam).
+    const shallowPropsEqual = shallowEqual(this.props, nextProps);
+
+    // We also need to check the imgixParams.
+    const shallowParamsEqual = shallowEqual(
+      this.props.imgixParams,
+      nextProps.imgixParams
+    );
+
+    return shallowPropsEqual && shallowParamsEqual;
   }
 
   render() {
--- a/src/react-imgix-bg.jsx
+++ b/src/react-imgix-bg.jsx
@@ -1,11 +1,10 @@
 import React from "react";
 import { withContentRect } from "react-measure";
-import { PACKAGE_VERSION } from './constants';
+import { PACKAGE_VERSION } from "./constants";
 import constructUrl from "./constructUrl";
 import extractQueryParams from "./extractQueryParams";
 import findClosest from "./findClosest";
 import targetWidths from "./targetWidths";
-
 
 const noop = () => {};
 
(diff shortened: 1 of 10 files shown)


src/react-imgix-bg.jsx
```diff
@@ -1,120 +1,155 @@
...
-    if (typeof ref === "function") {
-      ref(el);
-    }
-  };
+class BackgroundImpl extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  shouldComponentUpdate(nextProps) {
```
frederickfogerty, line 21: This also needs to handle cases when other props change, e.g. `imgixParams`, `disableLibraryParam`, etc.

test/integration/react-imgix.test.jsx
```diff
@@ -568,7 +568,61 @@ describe("Background Mode", () => {
...
+    // If the browser width shrinks, then the "w" we're
+    // getting should be greater than or equal to the
+    // expectedWidth (in this case "w" doesn't change at
+    // all).
+    expect(
+      parseInt(bgImageSrcURL.getQueryParamValue("w"))
+    ).toBeGreaterThanOrEqual(expectedWidth);
+  });
+
+  it("window resize larger", async () => {
```
frederickfogerty, line 581: I like these tests! 👏

frederickfogerty (changes requested):
This is looking good! It seems like the initial implementation really was quite broken. 

Something else I was wondering when reading this: does this change the re-rendering behaviour that is ultimately caused by react-measure? i.e. is `shouldComponentUpdate` continuously called? Or is this actually stopping react-measure from trying to re-render constantly?

ericdeansanchez:
>Something else I was wondering when reading this: does this change the re-rendering behaviour that is ultimately caused by react-measure? i.e. is shouldComponentUpdate continuously called? Or is this actually stopping react-measure from trying to re-render constantly?

The original behavior is re-rendering occurs from the moment of first-render onward. The component re-renders ~70 times per second (and there's nothing stopping it, this is also _"70 times per second without resizing the browser"_).

The new behavior renders 3 times on first render which is expected given how `onResize` works for the `Measure` component ([it is called twice on mount](https://github.com/souporserious/react-measure#onresize-proptypesfunc)).

`shouldComponentUpdate` is not called continuously. It is called when props have been updated. Specifically, the `contectRect` changes within the `WithContentRect` wrapper component, but a re-render only occurs if the width/height within the `contentRect` have increased.

Now we can count renders: `initial + 3 resizes smaller + 3 resizes larger = 7 measures = 21 renders` (and no more). If you were to try to collect a similar count on the original behavior it's: `7 measures = 485 renders` (more or less depending on how quickly you perform the resizes, again because it's constantly re-rendering.

Visually, this first image is the number of renders with the original component mounting for the first time only. No resizing occurs and I only recorded for 1.3s (e.g. 78th of 78 renders at 1.3s).

70-renders-per-second

This image is of how the updated component renders. I started out with a small browser window and upsized a few times. Here the component actually needs to update; it does so at the 13.4s mark. From this you can see the improvement over rendering 70x/sec.
last-recorded-update-to-contentRect

And here is the final update, but only to the underlying `contentRect`:
last-render

Now, there is room for improvement here. `contentRect` is a little noisy wrt to updates from the `ResizeObserver` `entry`. That's not a huge deal. Lower hanging fruit would be to only update if height has increased beyond some constant factor of the previous amount (i.e. re-rendering for a height change from 27px to 28px doesn't seem very ideal, but this is all meant to be an iterative start).

frederickfogerty:
Thanks for the in depth reply here @ericdeansanchez!

src/react-imgix-bg.jsx
```diff
@@ -13,107 +14,153 @@ const findNearestWidth = (actualWidth) =>
...
-    if (forcedWidth != null) {
-      const height = Math.ceil(forcedWidth / ar);
-      return { width: forcedWidth, height };
-    } else if (forcedHeight != null) {
-      const width = Math.ceil(forcedHeight * ar);
-      return { width, height: forcedHeight };
+
+    // If we made it here, we need to check if the "top-level"
+    // props have changed (e.g. disableLibraryParam).
+    const shallowPropsEqual = shallowEqual(this.props, nextProps);
```
sherwinski, line 52: This may be more than we eventually want to check for, but I think it's a good enough starting point for now 👍

sherwinski (approved):

src/react-imgix-bg.jsx
```diff
@@ -46,7 +47,17 @@ class BackgroundImpl extends React.Component {
...
+
+    // We also need to check the imgixParams.
+    const shallowParamsEqual = shallowEqual(
+      this.props.imgixParams,
+      nextProps.imgixParams
+    );
+
+    return shallowPropsEqual && shallowParamsEqual;
```
frederickfogerty, line 60: This is great so far! Two things that should be done here:

1. `imgixParams` shouldn't be checked on L52
2. `htmlAtrtibutes` should also be shallowly compared.

This could be done like this:
```jsx
    const customizer = (value, other, key) => {
        if (key === "imgixParams" || key === "htmlAttributes") {
            return shallowEqual(value, other);
        }
        return undefined; // Use default behaviour
    };

    // If we made it here, we need to check if the "top-level"
    // props have changed (e.g. disableLibraryParam).
    const shallowPropsEqual = shallowEqual(this.props, nextProps, customizer);

    return shallowPropsEqual;
```

PR #790, opened 2021-05-25
feat: create ixPropsReducer

luqven:
This PR creates:

- `mergeComponentPropsHOF()`: tries to invoke `useImgixContext()`. If
context is `undefined`, it's being accessed outside of `ImgixContext`
provider and the Component is returned as is. Otherwise, the Provider
and Child Component props are combined and passed down as the Child
Component's props.

- `processPropsHOF()`: formats a Component's props to adhere to imgix
  client's specifications.

This commit also creates and the `mergeProps()`, `formatProps()`,
and `formatSrc()` helper functions.

- formatSrc: creates a 1-step URL from `domain` and `src` Strings.
- formatProps: formats the following values in the props Object:
  - `width`: if negative gets set to `undefined`.
  - `height`: if negative gets set to `undefined`.
  - `src`: concatenated to `domain` if `src` defined and has no domain.
- mergeProps: Merges the `src` object into the `destination` object. 
Destination values are not overwritten by source values. Destination 
properties that resolve to `undefined` or `null` are not overwritten if
a destination value exists unless destination key does not exist .
It recursively merges the `imgixParams` and `htmlAttributes` values.
client's specifications.

## Remaining work

There's a non-trivial amount of URL formatting happening here that gives the whole thing a bad smell. Still iterating on how to best remove this necessity. Ideally, `@imgix/js-core` would have a `domain` parameter. Open to suggestions.

## Up for discussion

The js-doc comments are _very_ .. verbose. Please feel free to ruthlessly tear them apart if they're getting in the way of you understanding. They've been added in an effort to make reading this library easier for maintainers via the hover-tooltips in IDEs.

----
This PR is part of a stack

| PR 	| Title 	| Merges Into 	|   	|
|----	|-------	|-------------	|---	|
|    1	|    [imgixProvider](https://github.com/imgix/react-imgix/pull/791)   	|        [luis/PE-828](https://github.com/imgix/react-imgix/tree/luis/PE-828)    	|   	|
|    2	|    [propsMerger](https://github.com/imgix/react-imgix/pull/790)  👈🏼   	|           #791  	|   	|
|    3	|    [wrapReactImgix](https://github.com/imgix/react-imgix/pull/792)    	|       #790      	|   	|

> edit: removed old implementation details. Added new implementation details.


frederickfogerty (changes requested):
I think this is starting down the right track, but I think it could be a little bit better to separate the concerns here. Right now this a) merges the two props objects together, and b) does some processing of the resulting props.

I think it would be better to keep this function to just doing (a), and moving (b) closer to the component's render method, since then it's easier to grok where that processing is being done. Also it makes this function smaller and less error-prone. 

To be clear, imo (a) is anything where two or more props interact, e.g. combining `src` and `domain` into `src`, and (b) is transforming single props, e.g. `width`, `height`

---

> The reducer terminology can be argued to be misleading / confusing. This function is combining two objects. Open to suggestions on how to best refer to it.

I also agree, what about something like `propsMerger`?

src/HOFs/propFormatter.js
```diff
@@ -0,0 +1,79 @@
...
+  const height = !props.height || props.height <= 1 ? null : props.height
+  const src = props.src ? formatSrc(props.src, props.domain) : null
+
+  return Object.assign( {}, props, { width, height, src,} )
+}
+
+// TODO(luis): write me
+export const collapseImgixParams = (params) => {
+  return params
+}
```
luqven, line 63: Very much a WIP 😆

src/HOFs/propMerger.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Otherwise, it merges a Component's props with the `ImgixContext` props and 
+ * return a Component with the merged `props`.
+ * @param {React.Element <typeof Component} Component -  with defined `props`.
+ * @returns Component with merged `props`.
+ */
+export const mergeComponentPropsHOF = (Component) => (props) => {
+  const contextProps = useImgixContext();
+
+  if (contextProps === undefined) {
```
frederickfogerty, line 47: Better to use `== null` here just to be safe

src/HOFs/propMerger.js
```diff
@@ -0,0 +1,53 @@
...
+ * let src = {foo: 'bar'}
+ * let destination = {foo: 'baz', bar: 'baz'}
+ * let result = mergeProps(src, destination)
+ * result => {foo: 'bar', bar: 'baz'}
+ * 
+ */
+export function mergeProps(src = {}, destination = {}) {
+  // ensure the arguments are not modified
+  Object.freeze(src);
+  Object.freeze(destination);
```
frederickfogerty, line 22: I don't think these are necessary

src/HOFs/propMerger.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {Object} destination - The child component's props object
+ * @returns Object with the combined values from `src` & `destination` Objects
+ * @example
+ * let src = {foo: 'bar'}
+ * let destination = {foo: 'baz', bar: 'baz'}
+ * let result = mergeProps(src, destination)
+ * result => {foo: 'bar', bar: 'baz'}
+ * 
+ */
+export function mergeProps(src = {}, destination = {}) {
```
frederickfogerty, line 19: I think this function needs to also recursively merge at least `imgixParams` and `htmlAttributes`, i.e.

```js
mergeProps({ imgixParams: { fit: "fill" } }, { imgixParams: { ar: "2:1" } }); 
// => { imgixParams: { fit: "fill", ar: "2:1" } },
```

frederickfogerty (approved):
Overall looks like a good approach here!

src/HOFs/propMerger.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+export const mergeProps = (src, destination) => {
+  if (src == null && destination !== null) {
+    return destination
+  }
+  if (src !== null && destination == null) {
+    return src;
+  }
+  if (src == null && destination == null) {
+    return undefined
```
frederickfogerty, line 45: Maybe it's better to return `{}` here?

src/HOFs/propFormatter.js
```diff
@@ -0,0 +1,89 @@
...
+ * - `height`: if undefined or negative gets set to `null`.
+ * - `src`: concatenated to `domain` if `src` defined and has no domain.
+ *
+ * @param {Object} props 
+ * @returns A formatted `props` Object.
+ */
+export const formatProps = (props) => {
+  const width = !props.width || props.width <= 1 ? null : props.width
+  const height = !props.height || props.height <= 1 ? null : props.height
+  const src = props.src ? formatSrc(props.src, props.domain) : null
```
frederickfogerty, line 49: I like this, but I really try avoid using `null` unless necessary, and much prefer `undefined`. What do you think?

src/HOFs/propFormatter.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {React.Element<typeof Component>} Component - A react component with
+ * defined `props`.
+ * @returns A React Component who's `props` have been formatted and 
+ * `imgixParams` have been collapsed.
+ */
+export const processPropsHOF = (Component) => (props) => {
+  const formattedProps = formatProps(props)
+  const formattedImgixParas = collapseImgixParams(formattedProps.imgixParams)
+
+  return <Component {...formattedProps} imgixParams={formattedImgixParas} />
```
frederickfogerty, line 88:
```suggestion
  const formattedImgixParams = collapseImgixParams(formattedProps.imgixParams)

  return <Component {...formattedProps} imgixParams={formattedImgixParams} />
```

test/unit/propMerger.test.js
```diff
@@ -0,0 +1,63 @@
...
+      const result = mergeProps(src, destination);
+
+      expect(result).toEqual(
+        {
+          width: 101,
+          height: 201,
+          imgixParams: { ar: "1:2", dpr: 1 },
+          htmlAttributes: { styles: "width: 100" }
+        }
+      );
```
frederickfogerty, line 59: This is missing some edge cases atm
```suggestion
      const src = {
        imgixParams: { ar: "1:2", dpr: 2},
        htmlAttributes: { styles: "width: 50", alt: "src" }
      }
      const destination = {
        imgixParams: { dpr: 1, fit: "fill" },
        htmlAttributes: { styles: "width: 100", className: "destination" }
      }
      const result = mergeProps(src, destination);

      expect(result).toEqual(
        {
          imgixParams: { ar: "1:2", dpr: 1, fit: "fill" },
          htmlAttributes: { styles: "width: 100", className: "destination", alt: "src" }
        }
      );
```

test/unit/propMerger.test.js
```diff
@@ -0,0 +1,63 @@
...
+    () => {
+      const src = { width: 100, height: 200, imgixParams: { ar: "1:1" } }
+      const destination = { width: 101, height: null, imgixParams: { rot: 90 } }
+      const result = mergeProps(src, destination);
+
+      expect(result).toEqual(
+        { width: 101, height: 200, imgixParams: { ar: "1:1", rot: 90 } }
+      );
+    }
+  )
```
frederickfogerty, line 34: Actually I'm not sure I think this is a good idea. I think it's pretty standard behaviour to have the destination overwrite the src if set to `undefined`. i.e this:

```js
expect(mergeProps({ height: 100 }, {})).toEqual({ height: 100 })

expect(mergeProps({ height: 100 }, { height: undefined })).toEqual({ height: undefined });
// or
expect(mergeProps({ height: 100 }, { height: undefined })).toEqual({});
```

test/unit/propMerger.test.js
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  it('should not overwrite destination values with source values', () => {
+    const src = { width: 100, height: 200, imgixParams: {ar: "1:2", dpr: 2} }
+    const destination = { width: 101 }
+    const result = mergeProps(src, destination);
+
+    expect(result).toEqual(
+      { width: 101, height: 200, imgixParams: { ar: "1:2", dpr: 2 } }
+    )
```
frederickfogerty, line 21:
```suggestion
    const src = { width: 100 }
    const destination = { width: 101 }
    const result = mergeProps(src, destination);

    expect(result).toEqual(
      { width: 101 }
    )
```

Better to exclude data that is not related to the test, the same can be done for the test below

src/HOFs/propFormatter.js
```diff
@@ -0,0 +1,89 @@
...
+ *   - Then, it joins the two strings on a `/` character. IE, 
+ *    `strippedDomain + "/" + strippedSrc`.
+ *   - If `domain` String argument `null` or `undefined`, the function returns
+ *    the original `src` String.
+ *
+ * @param {String} src - URL that is either 1-step or 2-step
+ * @param {String} domain - Domain string, optional
+ * @returns 1-step, or complete, URL String. Ex, _assets.ix.net/foo/bar.jpg_
+ */
+export function formatSrc(src, domain, useHTTPS = true) {
```
frederickfogerty, line 21: Althought there is nothing wrong with this function at all, it does seem a bit silly to combine the path and domain into a full src just to be destructured later into `domain` and `path` again. I know the plan was not to change the internals of ReactImgix and I applaud this PR for doing that, but I wonder if it would be better just to rip the bandaid off and fix this problem properly. Another option is to wait until `@imgix/js-core` supports 1-step URLs and then just remove the destructuring behaviour inside ReactImgix then.

Ignore this if you have other plans already.

frederickfogerty (approved):
Overall looks really good!

test/unit/propMerger.test.js
luqven: This ms 💯 , good callout.

Something like...

```js
  const newProps = { ...destination };
  const newPropKeys = Object.keys(newProps);

  for (const [k, v] of Object.entries(src)) {
    if (newPropKeys.indexOf(k) == -1 && v !== null) {
      newProps[k] = v;
    }
```

This will both ensure we add new k, v pairs to the `destination` and allow `destination` to overwrite `src` k, v pairs when set as null or undefined.

luqven:
> Overall looks really good!

Thanks for the review ❤️  Addressed all your comments. Please let me know if any further changes are needed here.

Re: https://github.com/imgix/react-imgix/pull/790#discussion_r644667372

There are some pending changes to `js-core` that should allow us to remove the destructuring at a later date.

sherwinski (approved):
👌

src/HOFs/propFormatter.js
```diff
@@ -0,0 +1,89 @@
+import * as React from 'react'
+import { PARAMS_EXP_MAP } from "./constants"
+
+/**
+ * Creates a 1-step, or complete, URL from `domain` and `src` Strings.
+ *
+ * - First, the function checks if src has a defined `domain`. If it does, it 
+ * checks to see if `src` has a scheme, and prepends "http" or "https" as needed
```
sherwinski, line 8: > and prepends "http" or "https"

Is there any situation where "http" is prepended? Based on L32 it looks like it will always be "https"

src/HOFs/propMerger.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import { useImgixContext } from "../HOCs"
+
+/**
+ * Merges the `src` object into the `destination` object. Destination values are
+ * not overwritten by source values. Destination properties that resolve to
```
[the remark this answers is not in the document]
luqven: Ah darn thanks, the jsdoc is out of date here 🤦🏼

src/HOFs/propFormatter.js
luqven: Shoot nice catch. 

As of right now will always prepend `https` since the `useHttps` argument isn't being used.

Ideally, if the Provider and Consumer combined props have `useHttps` prop as `false`, this function would prepend `http`.

src/HOFs/propFormatter.js
```diff
@@ -44,9 +44,9 @@ export function formatSrc(src, domain, useHTTPS = true) {
  * @returns A formatted `props` Object.
  */
 export const formatProps = (props) => {
-  const width = !props.width || props.width <= 1 ? null : props.width
-  const height = !props.height || props.height <= 1 ? null : props.height
-  const src = props.src ? formatSrc(props.src, props.domain) : null
+  const width = !props.width || props.width <= 1 ? undefined : props.width
+  const height = !props.height || props.height <= 1 ? undefined : props.height
+  const src = props.src ? formatSrc(props.src, props.domain) : undefined
```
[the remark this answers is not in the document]
luqven: 💯  ty, I missed this

src/HOFs/propFormatter.js
luqven: Addressed in this commit https://github.com/imgix/react-imgix/pull/790/commits/311082c487c0e2b6ec20895edf53a95701fb1130

test/unit/propMerger.test.js
```diff
@@ -19,13 +19,13 @@ describe('mergeProps()', () => {
     expect(result).toEqual({ width: 101 })
   })
 
-  it('should overwrite destination that resolves to `null` if source exists',
+  it('should not overwrite destination resolving to `null` if source exists',
```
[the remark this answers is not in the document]
luqven: addressed in this commit https://github.com/imgix/react-imgix/pull/790/commits/a2fca9cdb6911fac43b562a857d379fda0e5fc16

src/HOFs/propFormatter.js
luqven: Add tests in this commit https://github.com/imgix/react-imgix/pull/790/commits/595600cfa5bccde51f3c721fcaea5945a647257a

sherwinski (approved):
LGTM

PR #791, opened 2021-05-25
feat: create imgixProvider component

luqven:
This PR is part of a stack

| PR 	| Title 	| Merges Into 	|   	|
|----	|-------	|-------------	|---	|
|    1	|    [imgixProvider](https://github.com/imgix/react-imgix/pull/791) 👈🏼  	|        [luis/PE-828](https://github.com/imgix/react-imgix/tree/luis/PE-828)    	|   	|
|    2	|    [propsMerger](https://github.com/imgix/react-imgix/pull/790)     	|           #791  	|   	|
|    3	|    [wrapReactImgix](https://github.com/imgix/react-imgix/pull/792)    	|       #790      	|   	|
----
This commit creates a Provider component, `<ImgixProvider />`. The
component uses the `React.createContext()` API to create a props context
that can be modified and passed on to its child components. This is
useful as it allows for HOF that can format and combine imgixProps
before they're passed on `<ReactImgix />`. In order to, for example,
support 2-part URLs, IE domain + path, we can use the Provider to pass
these props to a HOF props merger, format the 2 parts into one combined
URL and finally pass the combined props to `<ReactImgix />`.

This allows us to leave the current API for `<ReactImgix>` unchanged while
also unchanged while also adding support for new props now and in the
future.

For example, the ImgixProvider can store a domain prop that is shared
across multiple `<ReactImgix>` components, each with a distinct src prop.
Inside the Provider, the props reducer combined the shared domain prop
with the src props to create a one-step URL.

### DIFF
--- a/src/HOCs/imgixProvider.jsx
+++ b/src/HOCs/imgixProvider.jsx
@@ -3,10 +3,9 @@ import React, { useContext, createContext } from 'react'
 const ImgixContext = createContext()
 
 /**
- * `useImgixContext()` tries to invoke `React.useContext()`. If context is 
- * `undefined`, we know that the context is being accessed outside of a 
- * provider, and therefore throw an `Error`.
- * @returns Hook that gets context value from the nearest`ImgixProvider`.
+ * `useImgixContext()` tries to invoke `React.useContext()`. If no context
+ * is available, this function returns `undefined`. 
+ * @returns The context defined by the closest parent `ImgixProvider`.
  */
 function useImgixContext() {
   return useContext(ImgixContext)
@@ -37,4 +36,3 @@ function ImgixProvider({children, ...reactImgixProps}) {
 // }
 
 export {ImgixProvider, useImgixContext}
-
--- a/src/HOCs/imgixProvider.jsx
+++ b/src/HOCs/imgixProvider.jsx
@@ -9,11 +9,7 @@ const ImgixContext = createContext()
  * @returns Hook that gets context value from the nearest`ImgixProvider`.
  */
 function useImgixContext() {
-  const context = useContext(ImgixContext)
-  if (context === undefined) {
-    console.error("useImgixContext must be used within a ImgixProvider")
-  }
-  return context
+  return useContext(ImgixContext)
 }
 
 /**
@@ -27,6 +23,9 @@ function useImgixContext() {
 function ImgixProvider({children, ...reactImgixProps}) {
   const value = reactImgixProps
 
+  if ( children == null || children.length < 1) {
+    console.error("ImgixProvider must have at least one Imgix child component")
+  }
   return <ImgixContext.Provider value={value}>{children}</ImgixContext.Provider>
 }
 
--- a/src/HOCs/imgixProvider.jsx
+++ b/src/HOCs/imgixProvider.jsx
@@ -28,11 +28,4 @@ function ImgixProvider({children, ...reactImgixProps}) {
   return <ImgixContext.Provider value={value}>{children}</ImgixContext.Provider>
 }
 
-
-//TODO(luis): do we still need this?
-// const propsProcessorHOF = (Component) => (props) => {
-//   const collapsedImgixParams = collapseImgixParam(props.imgixParams);
-//   return <Component {...props} imgixParams={collapsedImgixParams} />
-// }
-
 export {ImgixProvider, useImgixContext}
--- a/src/HOCs/imgixProvider.jsx
+++ b/src/HOCs/imgixProvider.jsx
@@ -17,7 +17,7 @@ function useImgixContext() {
  * `React.useContext()` API.
  * @param {React.Element <typeof Component>} children 
  * @param {Object} reactImgixProps 
- * @returns Component
+ * @returns React.Element
  */
 function ImgixProvider({children, ...reactImgixProps}) {
   const value = reactImgixProps
--- a/src/HOCs/imgixProvider.jsx
+++ b/src/HOCs/imgixProvider.jsx
@@ -1,0 +1,41 @@
+import React, { useContext, createContext } from 'react'
+
+const ImgixContext = createContext()
+
+/**
+ * `useImgixContext()` tries to invoke `React.useContext()`. If context is 
+ * `undefined`, we know that the context is being accessed outside of a 
+ * provider, and therefore throw an `Error`.
+ * @returns Hook that gets context value from the nearest`ImgixProvider`.
+ */
+function useImgixContext() {
+  const context = useContext(ImgixContext)
+  if (context === undefined) {
+    console.error("useImgixContext must be used within a ImgixProvider")
+  }
+  return context
+}
+
+/**
+ * Creates a Provider component that passes `reactImgixProps` as the Context 
+ * for child components who use the `useImgixContext()` custom hook or 
+ * `React.useContext()` API.
+ * @param {React.Element <typeof Component>} children 
+ * @param {Object} reactImgixProps 
+ * @returns Component
+ */
+function ImgixProvider({children, ...reactImgixProps}) {
+  const value = reactImgixProps
+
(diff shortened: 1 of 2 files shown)


frederickfogerty (commented):
This is on the right track! Let me know when this is ready for a full review.

src/HOCs/imgixProvider.jsx
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * `useImgixContext()` tries to invoke `React.useContext()`. If context is 
+ * `undefined`, we know that the context is being accessed outside of a 
+ * provider, and therefore throw an `Error`.
+ * @returns Hook that gets context value from the nearest`ImgixProvider`.
+ */
+function useImgixContext() {
+  const context = useContext(ImgixContext)
+  if (context === undefined) {
+    console.error("useImgixContext must be used within a ImgixProvider")
```
frederickfogerty, line 14: This will always log an error when there is no context, right? This probably does more harm than good

frederickfogerty (approved):
Overall looks like a good approach here!

src/HOCs/imgixProvider.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useContext, createContext } from 'react'
+
+const ImgixContext = createContext()
+
+/**
+ * `useImgixContext()` tries to invoke `React.useContext()`. If context is 
+ * `undefined`, we know that the context is being accessed outside of a 
+ * provider, and therefore throw an `Error`.
+ * @returns Hook that gets context value from the nearest`ImgixProvider`.
```
frederickfogerty, line 9:
```suggestion
 * `useImgixContext()` tries to invoke `React.useContext()`. If no context
 * is available, this function returns `undefined`. 
 * @returns The context defined by the closest parent `ImgixProvider`.
```

I verified that `React.useContext` does not throw an error if context is not passed

src/HOCs/imgixProvider.jsx
```diff
@@ -0,0 +1,40 @@
...
+  return useContext(ImgixContext)
+}
+
+/**
+ * Creates a Provider component that passes `reactImgixProps` as the Context 
+ * for child components who use the `useImgixContext()` custom hook or 
+ * `React.useContext()` API.
+ * @param {React.Element <typeof Component>} children 
+ * @param {Object} reactImgixProps 
+ * @returns Component
```
frederickfogerty, line 21:
```suggestion
 * @returns React.Element
```

src/HOCs/imgixProvider.jsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+  return <ImgixContext.Provider value={value}>{children}</ImgixContext.Provider>
+}
+
+
+//TODO(luis): do we still need this?
+// const propsProcessorHOF = (Component) => (props) => {
+//   const collapsedImgixParams = collapseImgixParam(props.imgixParams);
+//   return <Component {...props} imgixParams={collapsedImgixParams} />
+// }
```
frederickfogerty, line 37: Maybe remove this?

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  test('should not have context value defined if Provider has no props', () => {
+    const wrappedComponent = (
+      <ImgixProvider>
+        <ReactImgix {...imageProps}/>
+      </ImgixProvider>
+    )
+
+    const expectedProps = {
+      children: < ReactImgix src="https://assets.imgix.net/examples/pione.jpg" sizes="50vw" />,
```
frederickfogerty, line 26:
```suggestion
      children: <ReactImgix src="https://assets.imgix.net/examples/pione.jpg" sizes="50vw" />,
```

Seems like this PR could use some Prettier TLC

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const imageProps = {
+  src: "https://assets.imgix.net/examples/pione.jpg",
+  sizes: "50vw",
+}
+
+describe('ImgixProvider', () => {
+
+  test('should not have context value defined if Provider has no props', () => {
```
frederickfogerty, line 18: So while these tests work, I think these tests are brittle because we're depending on the order of composition of the higher order components on `ReactImgix`. For example, I think if we add another higher order component or change the order around, these tests would break. A potential better solution would be a) traverse the tree to check props only on a specific component name, or b) use some sort of spy (I know we had issues with this). 

There's no change needed because it's too much effort to fix for the benefit, but something to keep in mind for next time round.

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+    const renderedComponent = shallow(wrappedComponent)
+    expect(renderedComponent.props()).toEqual(expectedProps)
+  })
+
+  // TODO(luis): enable these tests once propMerger and propProcessor merged in
+  // test('should merge the Provider and Child props', () => {
+
+  //   const modifiedProps = { ...imageProps };
+  //   modifiedProps.src = "examples/pione.jpg"
+  //   modifiedProps.sizes = null
```
frederickfogerty, line 56:
```suggestion
  //   const modifiedProps = { 
  //     ...imageProps,
  //     src: "examples/pione.jpg",
  //     sizes: null,
  //   };
```

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+  //   const renderedComponent = mount(wrappedComponent)
+  //   const renderedProps = renderedComponent
+  //     .childAt(0) // mergePropsHOF
+  //     .childAt(0) // processPropsHOF
+  //     .childAt(0) // ChildComponent
+  //     .props()
+  //   // remove noop function that breaks tests
+  //   renderedProps.onMounted = undefined
+
+  //   expect(renderedProps).toEqual(expectedProps)
```
frederickfogerty, line 85:
```suggestion
  //   expect(renderedProps).toMatchObject(expectedProps)
```

should do the trick

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+  // })
+
+  // test('should log error when has no consumers', () => {
+  //   jest.spyOn(global.console, 'error').mockImplementation(() => {})
```
frederickfogerty, line 89: Do we need to reset this spy after the test?

test/unit/imgix-provider.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+  //   )
+
+  //   const expectedProps = {
+  //     children: undefined,
+  //     value: providerProps
+  //   }
+
+  //   const renderedComponent = shallow(wrappedComponent)
+
+  //   expect(renderedComponent.props()).toEqual(expectedProps)
```
frederickfogerty, line 103: This assertion doesn't match the test name - either this assertion should be deleted or the test name should be updated

frederickfogerty (approved):
LGTM!

src/HOCs/imgixProvider.jsx
luqven: Ah thanks, this jsdoc got outdated since the context check was moved outside of this function.

test/unit/imgix-provider.test.jsx
luqven: Ya, idk what's happening with prettier in this repo for me 🤔. Apols for the messy formatting.

test/unit/imgix-provider.test.jsx
luqven: This test won't work until the next PR in the sack is merged in. Going to remove it from this PR and instead add it back in whenever dependencies have been merged in.

sherwinski (approved):

frederickfogerty (approved):
🔥

Issue #789, opened 2021-05-19
Children passed to background component do not update


JosephMarotte:
**Before you submit:**

- [x] Please search through the existing issues (both open AND closed) to see if your issue has been discussed before. Github issue search can be used for this: https://github.com/imgix/react-imgix/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please ensure the problem has been isolated and reduced. This link explains more: http://css-tricks.com/6263-reduced-test-cases/

**Describe the bug**
`Background` component do not update although I believe it should.

**To Reproduce**

Link to sandbox : https://codesandbox.io/s/musing-cray-w6tel?file=/src/App.js

Code sample :
```js
import { Background } from 'react-imgix';
import React, { useState } from 'react';
 
export default function App() {
  const [counter, setCounter] = useState(0);
  return (
    <>
    <Background 
    src="https://images.prismic.io/slicemachine-blank/30d6602b-c832-4379-90ef-100d32c5e4c6_selfie.png?auto=compress,format">
      {counter}
    </Background>
    {counter}
    <button onClick={() => setCounter(counter + 1)}>increment counter</button>
    </>
  );
}
```

Steps to reproduce the behaviour:

1.  Go to https://codesandbox.io/s/musing-cray-w6tel?file=/src/App.js
2.  Click on `Increment counter` button
3.  See error

**Expected behaviour**
React children props passed to `Background` component should update.

**Information:**

- react-imgix version: 9.1.1
- browser version: whatsmybrowser.org/b/ZKTOF

**Additional context**
I think the problem lies with the following lines of `react-img-bg.jsx` :
https://github.com/imgix/react-imgix/blob/main/src/react-imgix-bg.jsx#L50-L60

```js
    // If we made it here, we need to check if the "top-level"    
    // props have changed (e.g. disableLibraryParam).
     const shallowPropsEqual = shallowEqual(this.props, nextProps);
    // We also need to check the imgixParams.
    const shallowParamsEqual = shallowEqual(this.props.imgixParams, nextProps.imgixParams);
    return shallowPropsEqual && shallowParamsEqual;
 ```
 
 The return statement should be `!(shallowPropsEqual && shallowParamsEqual)`

frederickfogerty:
[...]

PR #793, opened 2021-05-27
fix(789): ensure children passed to background component update 

luqven:
The `shouldComponentUpdate` function is refactored outside of the `BackgroundImpl` component.

The refactored function, `__shouldComponentUpdate` takes `props` and `nextProps` as arguments. It returns `true` if the `children`, `imgixParams`, `contextRect`, `bounds`, `htmlAttributes`, or top-level `props` have changed. It returns `false` if the props have not changed.

The refactoring addresses an issue where: 
- changes to the children and imgixParams props were not detected
- the child components therefore did not re-render.

``` js
// src/react-imgix-bg.jsx  

// L15
export const __shouldComponentUpdate = (props, nextProps) => {
// ... choose which props to ignore and which ones to compare
  const customizer = (oldProp, newProp, key) => {
    // ... false if children prop has changed
    if (key === "children") {
      return oldProp == newProp;
    }
    // ... false if an imgixParam has changed
    if (key === "imgixParams") {
      return shallowEqual(oldProp, newProp, (a, b) => {
        if (Array.isArray(a)) {
          return shallowEqual(a, b);
        }
        return undefined;
      });
    }
  }
  // ... false if props have changed
  const propsEqual = shallowEqual(props, nextProps, customizer);

  return !(propsEqual);
}
```

This PR closes issue #789.

### DIFF
--- a/src/react-imgix-bg.jsx
+++ b/src/react-imgix-bg.jsx
@@ -7,12 +7,71 @@ import findClosest from "./findClosest";
 import targetWidths from "./targetWidths";
 import { shallowEqual } from "./common";
 
-const noop = () => {};
-
 const findNearestWidth = (actualWidth) =>
   findClosest(actualWidth, targetWidths);
 
 const toFixed = (dp, value) => +value.toFixed(dp);
+
+export const __shouldComponentUpdate = (props, nextProps) => {
+  const contentRect = props.contentRect;
+  const bounds = contentRect.bounds;
+  const { width: prevWidth, height: prevHeight } = bounds;
+
+  const nextContentRect = nextProps.contentRect;
+  const nextBounds = nextContentRect.bounds;
+  const { width: nextWidth, height: nextHeight } = nextBounds;
+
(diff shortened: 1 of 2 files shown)


frederickfogerty (approved):
LGTM!

src/react-imgix-bg.jsx
```diff
@@ -7,57 +7,79 @@ import findClosest from "./findClosest";
...
+  if (prevHeight && nextHeight && nextHeight > prevHeight) {
+    return true;
+  }
 
-    // If neither of the previous nor next dimensions are present,
-    // re-render.
-    if (!nextWidth || !nextHeight || !prevWidth || !prevHeight) {
+  const customizer = (oldProp, newProp, key) => {
+    // these keys are ignored from prop checking process
+    if (key === "contextRect" || key === "measure" || key === "measureRef") {
```
luqven, line 45: @luqven this is misspelled, should be `contentRect`. Therefore, this will always return undefined or false.

PR #798, opened 2021-06-08
build: bump package lock semver which is out of sync

luqven:
I introduced a test-suite failure a few commits back when I bumped this manually after semantic release took "too long". It broke things. ~Hopefully~ this fixes things.


frederickfogerty (approved):
LGTM

Issue #723, opened 2020-10-30
Generated srcsets cannot be controlled


tremby:
This is discussed in the closed PR #528. I think it deserves an open issue, since I see the consequences of this missing feature as a major bug.

I'll quote the comment I left on that a little earlier:

>At the moment react-imgix is putting entries in the srcset way up to 8192 pixels wide. If my image is only 1000 pixels wide, a browser on a display with a lot of pixels is going to be requesting images way bigger than the source, and Imgix is going to pointlessly upscale the image, wasting the user's time and data.
>
>I can mitigate this by adding `fit=max` so at least it's never upscaling, but the browser will still download what it thinks are going to be larger versions of the image if the viewport size is increased, say, from 1400px wide to 2000px wide. But for all I know browser implementations may then have unexpected behaviour when the received image isn't actually the width the srcset value said it was going to be.
>
>I should definitely be able to tell Imgix what I know to the the largest available width of the image so it doesn't falsely promise the user that larger versions are available.
>
>To me the lack of this option seems like a very important bug. It's potentially a lot of wasted CPU cycles on both the Imgix side and the client side, wasted cache space on both sides, wasted network traffic, and wasted transfer time.
>
>The lack of ability to make the gap larger between available sizes is another issue -- the smaller the gap is, the less likely it is that the size the browser decides to download has already been generated and cached, meaning a more sluggish experience on average.

From other comments on that PR it looks like the idea was to rewrite parts of this component to delegate to imgix-core-js.

sherwinski:
Hey @tremby,

> From other comments on that PR it looks like the idea was to rewrite parts of this component to delegate to imgix-core-js.

This is still our intention and is slated to be worked on in the near future! Our team has had to focus our efforts elsewhere this year but I'm happy to say we are finally getting around to adding these features. We don't have a timeline currently for when this will be done, but our team can give updates on this issue as they come along.

> The lack of ability to make the gap larger between available sizes is another issue -- the smaller the gap is, the less likely it is that the size the browser decides to download has already been generated and cached, meaning a more sluggish experience on average.

The eventual integration of imgix-core-js will allow users to customize this aspect of their `srcset` as well.

> I can mitigate this by adding `fit=max` so at least it's never upscaling, but the browser will still download what it thinks are going to be larger versions of the image if the viewport size is increased, say, from 1400px wide to 2000px wide. 

[...]

PR #803, opened 2021-07-08
feat: add support for customizing srcset generation

cecchi:
## Description

This PR closes https://github.com/imgix/react-imgix/issues/723.

This adds support for controlling how the `srcset` attribute is generated, exposing the options supported by [ImgixClient.buildSrcSet](https://github.com/imgix/js-core#imgixclientbuildsrcsetpath-params-options): 

 - `widths`
 - `widthTolerance`
 - `minWidth`
 - `maxWidth`

I prefixed each option's equivalent prop with `srcSet` to make the purpose clear. I did not add one for `disableVariableQuality` since that is already controlled by the `disableQualityByDPR` prop.

Our use-case is that the generated `srcset` attributes are larger and overly granular than what we reasonably need. We have ~zero users who need 5000px wide images, and few source images that are large enough to render effectively at that size anyway. The size of the generated HTML can add up: a single srcset attribute is roughly 4kb when using the default settings. When rendering potentially hundreds of images on a single page, the total size is not trivial. These repetitive strings gzip very efficiently, but it's still unnecessary markup we'd rather do without.

## Checklist

I have not written documentation or tests for this feature yet, but am happy to if there is not opposition to the addition of these props in the first place. This change is backwards compatible.

- [x] Update the readme
- [ ] Add some [steps](#steps-to-test) so we can test your cool new feature!
- [ ] Add your info to the [contributors](#packagejson-contributors) array in package.json!

## Steps to Test

I do not have a simple test repo, but this should be trivial to test by checking out the branch locally, building, and linking to any existing app using `react-imgix`. Usage looks like:

```tsx
<Image
  src="https://assets.imgix.net/unsplash/bear.jpg"
  srcSetOptions={{
   widthTolerance: 0.15,
   maxWidth: 3000
  }}
/>
```


sherwinski:
Thanks for the PR @cecchi! We're definitely in favor of surfacing these options to users but let me take this back to the team for some discussion and see if there are any strong opinions on the interface. I'll get back to you sometime in the next week on how we'd like to proceed.

cecchi:
Hi @sherwinski, any updates from your team?

sherwinski:
Hey @cecchi ,
After discussing with the team, the major feedback given was that we'd prefer to contain these options within an object such as `srcsetOptions` rather than individual top-level props. If you don't mind making that adjustment, then we can move forward with writing tests and docs as well. I'm happy to answer any questions or provide guidance as needed 👍

cecchi:
Thanks @sherwinski. I've updated the PR with the proposed API, and added tests and some documentation to the README. I opted to link to the `@imgix/js-core` docs rather than redocument the options here, but feel free to update the docs however you feel is best.

frederickfogerty:
Hey @cecchi, thanks for the PR here! There are a few things to be changed before merging but overall this looks good, so I've gone ahead and done that in #807 since I couldn't push any changes to your fork/this PR. That PR will merge once the tests pass 👍

PR #807, opened 2021-08-05
feat: add support for customizing srcset generation

frederickfogerty:
This PR supersedes #803


sherwinski (approved):
LGTM, thanks @cecchi and @frederickfogerty 👍

cecchi:
An anecdotal follow-up here: we tweaked the settings in our implementation to use `{ widthTolerance: 0.2, maxWidth: 5000 }` and have been happy with the results. On some image-heavy pages this saved us 200kb+ of HTML served over the wire (much less after gzip, but still significant). Viewing the source of a page like [this one](https://www.curated.com/c/skiing?_skipHydration=true&_skipApolloCacheDump=true) should make it apparent why this was such a win for our use-case. I haven't dug enough into the data yet, and we are likely paying the some cost on the other end by requesting slightly larger images, but the TTFB win is important to us. In any event, sharing because it may be worth considering loosening Imgix's defaults for all users 🤷

frederickfogerty:
Thanks for that info @cecchi. We're glad to hear that it helped your page performance. As for imgix defaults, we'd have to do a more thorough investigation into a wider variety of pages.

PR #849, opened 2022-02-24
feat: allow path encoding to be disabled

### LINKED ISSUE #815: Version 9 over-encodes file paths
**Describe the bug**
When given a `src` which already has uri-encoded elements, the library is encoding them a second time, breaking the image path.

**To Reproduce**
Consider a file name saved with spaces: `spaced file.jpg`, which has been saved by/into a cms and encoded to display properly as `spaced%20file.jpg`. When `spaced%20file.jpg` is provided to _react-imgix_, the output shows this percent symbol as encoded, yielding a path of `spaced%2520file.jpg` which does not exist.

**Expected behaviour**
Filenames are not mis-encoded and return an appropriate source.

**Information:**
Downgrading back to `8.6.4` resolves the issue. It appears, though I may be mistaken, that this was introduced in [#780](https://github.com/imgix/react-imgix/pull/780) and the upstream dependency in core-js's [_sanitizePath](https://github.com/imgix/js-core/blob/main/src/index.js#L96) method.

- react-imgix version: Discovered on `9.2.0`
- browser version: `FF DE 93.0b4`

frederickfogerty:
This PR exposes the feature to disable path encoding, that was recently added to js-core. This behaviour is exposed through a new `disablePathEncoding` prop. Information about this new prop can be found in the README. 

Closes #815

### DIFF
--- a/README.md
+++ b/README.md
@@ -661,6 +661,10 @@ Disable generation of variable width src sets to enable responsiveness.
 
 By default this component adds a parameter to the generated url to help imgix with analytics and support for this library. This can be disabled by setting this prop to `true`.
 
+##### disablePathEncoding :: bool
+
+By default this component encodes the url path in the src and srcSet. This can be disabled by setting this prop to `true`. For more information about how imgix path encoding works, please refer to the [imgix/js-core](https://github.com/imgix/js-core#disable-path-encoding) docs.
+
 ##### htmlAttributes :: object
 
 Any other attributes to add to the html node (example: `alt`, `data-*`, `className`).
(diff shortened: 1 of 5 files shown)


frederickfogerty:
* main
  * **PR #849** <a href="https://app.graphite.dev/github/pr/imgix/react-imgix/849" target="_blank">Graphite</a>  (this PR)

sherwinski (changes requested):
Tracking some follow-up items before I forget:
1. We need to update js-core to enable the disable encoding feature
2. As per [this comment](https://github.com/imgix/react-imgix/issues/815#issuecomment-1050181249) we should investigate if this works with the provider component

frederickfogerty (commented):
@sherwinski I believe I've handled both of those concerns

sherwinski (approved):
LGTM

PR #850, opened 2022-02-25
fix: bug where _inPicture would be set on <img>'s that were children of <Picture>

frederickfogerty:
This was causing a React error during render


frederickfogerty:
* main
  * **PR #850** <a href="https://app.graphite.dev/github/pr/imgix/react-imgix/850" target="_blank">Graphite</a>  (this PR)

sherwinski (approved):
I'm a bit unfamiliar with this, are we expecting `_inPicture` to only be set on Source components then?

PR #857, opened 2022-03-07
chore(deps): move prop-types to dependencies

### LINKED ISSUE #843: prop-types should be a direct dependency
First of all, I love your service!

I believe `prop-types` should be a direct dependency instead of a peer dependency. 

We use TypeScript and in TS you don't really use `prop-types` so it's a bit awkward to explicitly install `prop-types` in our project just to satisfy `react-imgix`. It would make more sense if `react-imgix` had a dependency to what ever version of `prop-types` is needed to run `react-imgix`.

What are your thoughts on this?

sherwinski:
TSIA

Resolves #843

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,16 @@
 {
   "name": "react-imgix",
-  "version": "9.4.0",
+  "version": "9.5.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "react-imgix",
-      "version": "9.4.0",
+      "version": "9.5.0",
       "license": "ISC",
       "dependencies": {
         "@imgix/js-core": "^3.1.3",
+        "prop-types": "^15.8.1",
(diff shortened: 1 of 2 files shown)


sherwinski:
* main
  * **PR #857** <a href="https://app.graphite.dev/github/pr/imgix/react-imgix/857" target="_blank">Graphite</a>  (this PR)

package-lock.json
```diff
@@ -1,15 +1,16 @@
 {
   "name": "react-imgix",
-  "version": "9.4.0",
+  "version": "9.5.0",
```
sherwinski, line 3: Thought it was worth calling out that the lock file was out of sync before this.

frederickfogerty (approved):
LGTM

frederickfogerty:
By the way, this should probably have been a `fix` commit so that a release would have been cut

sherwinski:
will I ever learn

PR #876, opened 2022-09-26
ci(build): add build test to circleci steps

atlawrie:
Adds an additional test step to our CircleCI process to test successful builds.

# Before

Problems with builds can only be detected by CircleCI during the `deploy` step; this step only runs on `main` and prerelease branches. As a result, commits that break builds (included automated dependency updates by Renovate) in many cases might not be detected until after merge.
# After

We now test for a successful `build` step on every branch.


luqven (approved):
LGTM
check-checkmark-gif-by-most-expensivest

sherwinski (approved):

Issue #891, opened 2022-12-22
Width and height HTML attributes with fluid image


tremby:
**Before you submit:**

- [x] Please read the [contributing guidelines](CONTRIBUTING.md)
- [x] Please search through the existing issues (both open AND closed) to see if your feature has already been discussed. Github issue search can be used for this: https://github.com/imgix/vue/issues?utf8=%E2%9C%93&q=is%3Aissue
- [x] Please take a moment to find out whether your idea fits with the scope and aims of the project

**Is your feature request related to a problem? Please describe.**

[It's best practice to specify width and height on images](https://web.dev/optimize-cls/), even if they are fluid, to help avoid layout shift. It tells the browser the expected size before it loads any image data, and an aspect ratio is derived from it so the browser can still know the image size even if CSS or something else is altering the image's width from its intrinsic one.

But I cannot find a way to allow width and height attributes to get through to the browser when using the react-imgix component while leaving the image fluid width. As soon as I specify width and height, react-imgix assumes I want a fixed-width image, and it gives resolution descriptors in the srcset instead of width descriptors (and then my sizes attribute is ignored by the browser, as per spec).

**Describe the solution you'd like**

There should be a way to let imgix know the width and height of the image so it can pass it through to the browser, without signalling that you want a fixed size image. Perhaps allowing them to be added to `htmlAttributes` would be simplest.

**Describe alternatives you've considered**

I've tried adding width and height properties to `htmlAttributes` but they don't make it to the browser.

The only other alternative I can think of is not to use this component.

tremby:
Related, I saw some recent-ish patches removing width and height from `<source>` elements in `<picture>`. [Those are now allowed (and encouraged), however](https://chromestatus.com/feature/5737185317748736).

luqven:
Hey @tremby, thanks for opening this issue.

I toyed with this for a bit, and I agree that we should be able to pass down `width, height` via the `htmlAttributes` prop when they're not set on the component's `width` and `height` props. 

You're suggestion, LGTM. If you are okay with opening a PR, that'd be great, and we'd be happy to give you quick and constructive feedback.

Otherwise, we'll tackle this sometime next year.

Thanks!

tremby:
Another option would be a flag to explicitly tell the component if the user wants width descriptors or resolution descriptors. Which route seems best to you?

luqven:
That's interesting! Can you tell me more about how you'd want that to look/work?

[...]

PR #892, opened 2022-12-24
Allow width and height to be passed through htmlAttributes

tremby:
## Description

Before this PR there was no way to allow width and height attributes through to the `img` element while still using a srcset with width descriptors rather than resolution descriptors. Telling the browser the intrinsic dimensions and therefore the aspect ratio is important to avoid layout shift.

After this PR width and height are allowed through via the `htmlAttributes` prop, without affecting the logic Imgix takes to decide on srcset type.

To test, do something like

`<Imgix src="..." sizes="50vw" htmlAttributes={{width: 400, height: 200}} />` and the output should include the width and height, such as ``. Now the browser knows the aspect before any image data is loaded, so it can reserve the correct amount of space.

This PR also allows width and height attributes on `<source>` elements again, since they are now allowed by spec and encouraged.

## Checklist

- [x] Read the [contributing guidelines](CONTRIBUTING.md).
- [x] Each commit follows the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) spec format.
- [x] Update the readme (if applicable).
- [ ] Update or add any necessary API documentation (if applicable) -- not applicable
- [x] All existing unit tests are still passing (if applicable). -- yes, and new tests added, but the browser tests do not want to run so they are untested

- [x] Add some [steps](#steps-to-test) so we can test your bug fix or feature (if applicable).
- [x] Add new passing unit tests to cover the code introduced by your PR (if applicable).
- [ ] Any breaking changes are specified on the commit on which they are introduced with `BREAKING CHANGE` in the body of the commit. -- no breaking changes
- [ ] If this is a big feature with breaking changes, consider opening an issue to discuss first. This is completely up to you, but please keep in mind that your PR might not be accepted. -- not applicable

---

Commit message follows.

---

[It's best practice to specify width and height on images][1], even if they are fluid, to help avoid layout shift. It tells the browser the expected size before it loads any image data, and an aspect ratio is derived from it so the browser can still know the image size even if CSS or something else is altering the image's width from its intrinsic one.

Before this patch, there was not a way to allow width and height attributes to get through to the `img` element sent to the browser when using the react-imgix component while leaving the image fluid width (that is, with a srcset containing width descriptors). As soon as one specified width and height, react-imgix assumed the author wanted a fixed-width image, and it gave resolution descriptors in the srcset instead of width descriptors (and then the sizes attribute was ignored by the browser, [as per spec][2]).

Imgix currently uses the width and height attributes to decide whether to use width or resolution descriptors, so to avoid altering that logic (which would be a breaking change), this patch allows width and height to be passed through the `htmlAttributes` prop. Prior to the patch they were swallowed if given in `htmlAttributes`. As of this patch they are allowed and passed through.

This closes https://github.com/imgix/react-imgix/issues/891

In future it may make sense to decide srcset type based on the presence of the `sizes` attribute, as discussed in the above ticket. This would be a breaking change, however.

As part of this patch, width and height are also allowed through when using this component in `Source` mode, since [width and height attributes are now allowed and encouraged in that context][3].

[1]: https://web.dev/optimize-cls/
[2]: https://html.spec.whatwg.org/multipage/images.html#srcset-attributes
[3]: https://github.com/whatwg/html/pull/5894

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ This will generate HTML similar to the following:
 
 Since imgix can generate as many derivative resolutions as needed, react-imgix calculates them programmatically, using the dimensions you specify. All of this information has been placed into the srcset and sizes attributes.
 
-**Width and height known:** If the width and height are known beforehand, it is recommended that they are set explicitly:
+**Width and height known and fixed:** If the width and height are known beforehand, and a fixed-size image is wanted, it is recommended that they are set explicitly:
 
 ```js
 import Imgix from "react-imgix";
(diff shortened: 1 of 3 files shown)


tremby:
- CircleCI failure isn't to do with this commit I think; that looks like a CI config issue
- Is the other one failing because of a capital letter...?

tremby:
I guess not. I don't know what it's moaning about.

luqven:
Hey @tremby,

Apols for the delay in merging this in. We had some other projects that demanded all of our attention.

We'll get this out soon, thanks again for your hard work on this.

README.md
```diff
@@ -127,17 +127,44 @@ import Imgix from "react-imgix";
...
+
+**Width and height known but fluid:** If the image's intrinsic width and height are known but a fluid size image is wanted, [width and height should still be set to avoid layout shift](https://web.dev/optimize-cls/), but they must be set via `htmlAttributes` so as not to hint to Imgix to produce resolution descriptors in the srcset.
+
+```js
+import Imgix from "react-imgix";
+
+<Imgix
+  src="https://assets.imgix.net/examples/pione.jpg"
+  sizes="(min-width: 1024px) 40vw, 90vw"
+  htmlAttributes={{ // These are ignored by Imgix but passed through to the <img> element
```
luqven, line 140: nit(non-blocking): imgix should never be capitalized

```suggestion
  htmlAttributes={{ // These are ignored by imgix but passed through to the <img> element
```

README.md
```diff
@@ -127,17 +127,44 @@ import Imgix from "react-imgix";
...
+    width: 200,
+    height: 100,
+  }}
+/>;
+```
+
+In this example, Imgix will produce a srcset with width descriptors.
```
luqven, line 147:
```suggestion
In this example, imgix will produce a srcset with width descriptors.
```

README.md
```diff
@@ -127,17 +127,44 @@ import Imgix from "react-imgix";
 
 [![Edit xp0348lv0z](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/charming-keller-kjnsq)
 
+When width and height are specified, Imgix will give the image a srcset with resolution descriptors.
```
luqven, line 130:
```suggestion
When width and height are specified, imgix will give the image a srcset with resolution descriptors.
```

luqven (approved):
Thanks again for opening this PR, LGTM 👍🏼 

Just leaving some nits in here for future reference. Doesn't need to bock merging this PR in, but should get addressed in the future.

